PR #7, opened 2019-10-22
Feature/fix order of dispose

HenriBeck:
@tp I just noticed that we are not disposing the bloc when we create a new one because of changed inputs.

What is tricky here is that we can't just call `dispose` on it because then the `removeListener` in the `ValueListenableBuilder` would fail. Any ideas around the issue? We could save the old bloc instance and then potential call it in the `builder` method but this seems more like a hack too me.

I have added a test that makes sure that we create the correct amount of instances and then also dispose them correctly which currently fails because of that.

### DIFF
--- a/packages/with_bloc/lib/src/with_bloc.dart
+++ b/packages/with_bloc/lib/src/with_bloc.dart
@@ -45,6 +45,8 @@ class WithBlocState<BlocType extends ValueNotifier<StateType>, StateType>
   @visibleForTesting
   BlocType bloc;
 
+  BlocType _oldBloc;
+
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
@@ -57,7 +59,8 @@ class WithBlocState<BlocType extends ValueNotifier<StateType>, StateType>
     super.didUpdateWidget(oldWidget);
 
     if (!listId.equals(oldWidget.inputs, widget.inputs)) {
-      bloc.dispose();
+      /// Save the previous bloc so we can properly dispose it
+      _oldBloc = bloc;
 
       /// Recreate the bloc
       bloc = widget.createBloc(context);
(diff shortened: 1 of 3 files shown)


HenriBeck:
For now, I did the fix inside the `builder` method though that also looks like a hack to me, to be honest.

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -55,26 +59,37 @@ class WithBlocState<BlocType extends ValueNotifier<StateType>, StateType>
...
   Widget build(BuildContext context) {
     return ValueListenableBuilder<StateType>(
       valueListenable: bloc,
       child: widget.child,
-      builder: (context, value, child) =>
-          widget.builder(context, bloc, value, child),
+      builder: (context, value, child) {
+        /// We need to dispose the old bloc
+        if (_oldBloc != null) {
+          /// The old bloc needs to be disposed after [ValueListenableBuilder] had a change to update itself
```
Mrtcndkn: @HenriBeck Yeah it kinda looks hacky is there any performance or safety gain with the fix and didn't quite get the issue with "What is tricky here is that we can't just call dispose on it because then the removeListener in the ValueListenableBuilder would fail." can you elaborate a bit more?

tp:
>  Any ideas around the issue?

I think we could just drop the `ValueListenableBuilder` and just use `bloc.addListener(setState)`.

That way we could clean everything up ourselves when changing the blocs.

HenriBeck:
> > Any ideas around the issue?
> 
> I think we could just drop the `ValueListenableBuilder` and just use `bloc.addListener(setState)`.
> 
> That way, we could clean everything up ourselves when changing the blocs.

Right, that would make it easier.

packages/with_bloc/lib/src/with_bloc.dart
tp: @Mrtcndkn The issue with disposing the old bloc when creating the new bloc (in `didUpdateWidget`) would be that the `ValueListenableBuilder` will try to unsubscribe (`removeListener`) when it is updated and sees that the bloc reference has changed.

But by that point the old bloc would've already been disposed in which case `removeListener` would throw an error. That's why it's currently inside the inner `Builder`, which executes after the `ValueListenableBuilder` has unsubscribed and hence won't create any issue.

See https://github.com/flutter/flutter/blob/1aedbb1835bd6eb44550293d57d4d124f19901f0/packages/flutter/lib/src/widgets/value_listenable_builder.dart#L166

HenriBeck:
@tp I have removed our usage of `ValueListenableBuilder`

PR #8, opened 2019-10-22
package clean ups

tp:


packages/with_bloc/pubspec.yaml
```diff
@@ -12,7 +15,11 @@ dependencies:
   flutter:
     sdk: flutter
 
-  equatable: ^0.5.0
+  equatable: ^0.6.0
+
+  collection: ^1.14.0
```
HenriBeck: I think we could even drop these and write our own quick helper function

PR #10, opened 2019-10-22
Add "full" example

tp:
This shouldn't be our only example, just wanted to create a fuller example.

It's still WIP as I want to see how we could split out the UI and the actual login.

Also, even though nothing bad would happen if not implemented, I wanted to see how to add some "disable button immediately" functionality would look. Not sure yet if the below approach is even guaranteed in Flutter to guard against that in all circumstances. For all practical purposes it's probably sufficient to just do it based on the bloc's state.

### DIFF
--- a/packages/with_bloc/example/main.dart
+++ b/packages/with_bloc/example/main.dart
@@ -50,7 +50,7 @@ class LoginBloc extends StateQueue<LoginState> {
 class App extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
-    WithBloc<LoginBloc, LoginState>(
+    return WithBloc<LoginBloc, LoginState>(
       createBloc: (context) => LoginBloc(),
       builder: (context, bloc, state, _) {
         if (state is LoggedInState) {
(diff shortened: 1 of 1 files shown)


packages/with_bloc/example/main.dart
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    return completer.future;
+  }
+}
+
+class App extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    WithBloc<LoginBloc, LoginState>(
```
HenriBeck: Shouldn't we `return` here?

packages/with_bloc/example/main.dart
```diff
@@ -0,0 +1,146 @@
...
+          obscureText: true,
+          decoration: InputDecoration(
+            border: OutlineInputBorder(),
+            labelText: 'Username',
+          ),
+        ),
+        MaterialButton(
+          child: Text('Login'),
+          onPressed: !_loginInProgress
+              ? () async {
```
HenriBeck: Just for proper measures, I would make this a class method not to have inline methods inside the build function

HenriBeck:
@tp Do you know if the example needs to be a full flutter app? I thought that's what's being generated when you do `flutter create package`

tp:
> Do you know if the example needs to be a full flutter app?

No it can just be a sample file, not sure if it even gets compiled. 

But I guess having a proper, startle Flutter app can make sense, in which case one is supposed to add a `README.md` to the example directory and link to the individual examples oneself: https://dart.dev/tools/pub/package-layout#examples

packages/with_bloc/example/main.dart
tp: Absolutely!

packages/with_bloc/example/main.dart
tp: Done.

PR #12, opened 2019-10-22
Create inline docs for WithBloc

HenriBeck:


packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -1,30 +1,53 @@
...
     StateType value,
     Widget child,
   ) builder;
 
+  /// A [BlocType]-independent widget which is passed back to the [builder].
+  ///
+  /// This argument is optional and can be null if the entire widget subtree
+  /// the [builder] builds depends on the value of the [BlocType].
+  ///
+  /// Can be `null`.
```
tp: Is this redundant on purpose, or was it left over when switching from our usual nullability comments to the more elaborate one above?

HenriBeck: Yes, this was duplicate. I first added the `Can be null` and then added the more descriptive text

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -1,35 +1,53 @@
...
   final Widget Function(
     BuildContext context,
     BlocType bloc,
     StateType value,
     Widget child,
   ) builder;
 
+  /// A [BlocType]-independent widget which is passed back to the [builder].
+  ///
+  /// This argument is optional and can be null if the entire widget subtree
```
ahoseinian:
```
/// This argument is optional and can be `null`.
```
and what does `if the entire widget subtree the [builder] builds depends on the value of the [BlocType].` means ?

I couldn't follow the comment very well. do we pass the child as a performance optimisation ? so the child would not rebuild on each `[builder]` execution ?

HenriBeck: Yes, if you have a part of the subtree which is independent of the bloc and the value this can be a small optimization.

HenriBeck: https://github.com/aboutyou/dart_packages/pull/14/commits/f1b90ae220119eeea543dabf53e87140c3afda7e

ahoseinian: I am not sure if it is how you structure a sentence but with commas I can read it now.


also 
```suggestion
  ///
  /// The benefit of passing a widget down as a `child` is that it does 
  /// not get recreated on every `[builder]` execution. 
  ///
  /// This argument is optional and can be `null` if the entire widget subtree,
  /// which the [builder] builds, depends on the value of the [BlocType].

```

ahoseinian: doesn't matter with your new changes 👍

Issue #11, opened 2019-10-22
drop `collection` dependency 


tp:
I think we could even drop these and write our own quick helper function

_Originally posted by @HenriBeck in https://github.com/aboutyou/dart_packages/pull/8_

HenriBeck:
Fixed in #14

PR #14, opened 2019-10-23
Feature/remove collection dependency

HenriBeck:
#11


packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -41,8 +39,11 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
 
   /// A [BlocType]-independent widget which is passed back to the [builder].
   ///
-  /// This argument is optional and can be null if the entire widget subtree
-  /// the [builder] builds depends on the value of the [BlocType].
+  /// If a part of your subtree doesn't depend on the [BlocType] or [StateType],
+  /// it can be passed the [child] to this widget.
+  /// You will then get this [child] back in the [builder] method as the fourth argument.
+  /// Doing this will optimize the building of widgets, because whenever the [BlocType] changes,
+  /// it will not be recreated.
```
HenriBeck: Update the docs of the `child` argument here

ahoseinian: 👍

packages/with_bloc/test/are_lists_equal_test.dart
```diff
@@ -0,0 +1,79 @@
+import 'package:test/test.dart';
+import 'package:with_bloc/src/are_lists_equal.dart';
+
+class _Equatable {
```
tp:
```suggestion
/// Class whose instances are always equal to each other.
class _Equatable {
```

PR #18, opened 2019-11-26
Feature/implement pending operations for state queue

HenriBeck:
Builds on top of #17


packages/state_queue/lib/src/state_queue.dart
```diff
@@ -169,15 +178,22 @@ abstract class StateQueue<T> extends ValueNotifier<T>
   /// (neither directly nor indirectly), as that makes testing hard as the completion
   /// of the bloc can not be awaited properly.
   @protected
-  void run(StateUpdater<T> updater) {
-    _taskQueue.sink.add(_UpdaterEntry(updater));
+  Future<void> run(StateUpdater<T> updater) async {
```
tp: I have been thinking about adding a way to signal to the outside that an operation is done.

But I am not sure a `Future<void>` implemented this way is useful, as this always completes, not matter if there was an error, or if the action was ignored (due to an unexpected state type (e.g. loading) for example). One thing I was pondering with was returning the last state…

Should we maybe for now keep this as an implementation detail to get the pending operations working and restore the old `run` signature?

Or is there any new use-case that would desire this behavior, and not a full blown completer that also reports exceptions?

HenriBeck: Yeah sounds good. Then we could also talk about how that API should behave

packages/pending_operations/lib/src/pending_operations.dart
```diff
@@ -9,7 +9,7 @@ class PendingOperations implements PendingOperationsReader {
 
   int _pendingCalls = 0;
 
-  Future<void> _knownCallsHandled = Future<void>.value();
+  List<Future<void>> _knownCallsHandled = [];
```
HenriBeck: Previously this actually got stuck for me. Now we are only tracking the futures and then use `Future.wait` only in `awaitEndOfKnownCalls `

tp: Hmm, but the evaluation shouldn't change, `awaitEndOfKnownCalls` still has to be called in `tester.runAsync`. I just thought its most efficient to only keep track of the last future instead of all (not sure what else is kept alive by this.)

Will look into this in the overall scope again.

HenriBeck: As discussed: All of the futures are kept alive. It's just different in how we store them. Now we store in a flat list as before we had a nested list of `n` with always two items inside so we potentially had the overhead of `n * Futures.wait` in my opinion.

packages/state_queue/lib/src/state_queue.dart
```diff
@@ -94,6 +99,10 @@ abstract class StateQueue<T> extends ValueNotifier<T>
 
   final _taskQueue = StreamController<_QueueEntry<T>>();
 
+  final _pendingOperations = PendingOperations();
+
+  PendingOperationsReader get pendingOperations => _pendingOperations;
```
tp: As discussed, we might want to implement `runQueuedTasksToCompletion` using `pendingOperations.awaitEndOfKnownCalls`. Can also do that as a follow up.

tp:
Just for reference, the purpose of this change is to find out if a bloc has any ongoing (and possibly future) tasks scheduled. Otherwise we don't know when to stop calling `runToCompletion` in tests.

packages/pending_operations/lib/src/pending_operations.dart
tp: Guess it's fine for now as `awaitEndOfKnownCalls` is only used for testing; but I still want to look into it in the future, there should be a cleaner way that doesn't grow memory.

packages/state_queue/lib/src/state_queue.dart
```diff
@@ -176,14 +185,19 @@ abstract class StateQueue<T> extends ValueNotifier<T>
   /// of the bloc can not be awaited properly.
   @protected
   void run(StateUpdater<T> updater) {
-    _taskQueue.sink.add(_UpdaterEntry(updater));
+    final entry = _UpdaterEntry(updater);
```
tp: How about passing `done` to the `_UpdaterEntry` so we don't create the extra completer?

packages/state_queue/test/state_queue/state_queue_test.dart
```diff
@@ -52,4 +52,26 @@ void main() {
...
+  test(
+    'PendingOperations: should register operations for run method',
+    () async {
+      final bloc = _TestBloc();
+
+      bloc.setState(5);
+
+      expect(bloc.pendingOperations.pendingCalls, 1);
+
+      bloc.runQueuedTasksToCompletion();
```
tp: `unawaited(…)` ?

packages/state_queue/test/state_queue/state_queue_test.dart
```diff
@@ -52,4 +52,26 @@ void main() {
...
+
+      /// Will be still `1` as the [Timer.run] didn't run yet and so for has not marked the previous operation as done
```
tp:
```suggestion
      /// Will be still `1` as the [Timer.run] inside the `PendingOperations` didn't run yet and so far has not marked the previous operation as done
```

packages/state_queue/lib/src/state_queue.dart
HenriBeck: Done

packages/state_queue/test/state_queue/state_queue_test.dart
HenriBeck: No, it should be awaited here

packages/state_queue/test/state_queue/state_queue_test.dart
```diff
@@ -7,13 +7,13 @@ import 'package:test/test.dart';
 class _TestBloc extends StateQueue<int> {
   _TestBloc() : super(0);
 
-  void setState(int n) {
+  void setState(int n) async {
```
tp: Why `async` when they just return a `void`? Shouldn't change anything here.

tp: (Just makes me think that we should copy over the analyze configuration)

HenriBeck: Yeah let's do it after #20

PR #22, opened 2019-12-19
Create initial bloc in init state

HenriBeck:
My reasoning behind this is that we use a lot of `Provider.of<>(context)` in the `createBloc` without `listen: false`.

This might create a wrong assumption that in the case that these providers update, the bloc will be recreated. I think the proper way then is to move the `Provider.of(context)` out to the actual `build` method where you can actually listen to the changes.

This would be considered a breaking change.


tp:
> This might create a wrong assumption that in the case that these providers update, the bloc will be recreated. I think the proper way then is to move the Provider.of(context)

Doesn't `Provider.of(context, listen: false)` usually fall into 1 of 2 categories (in our app):

- Getting a global non-bloc that really is not changing throughout the app (API clients etc.)
- Getting a bloc/`ValueNotifier` (e.g. basket), so consumers can subscribe on that element directly instead of relying on the Flutter widget `build` to update them.


But I'll check out the code, maybe there's more to this change than I think of from the description.

HenriBeck:
@tp please have a quick look tomorrow here. It would be cool to get this released tomorrow

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -22,6 +22,13 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
   ///
   /// This function might be called multiple times in the lifetime of this widget.
   /// Whenever the [inputs] change, a new BLoC will be created by calling this function.
+  ///
+  /// For creating the initial bloc, this function will be called in the `initState`.
+  /// That means that you can listen to any `Provider`s or `MediaQuery`.
+  ///
+  /// If you need to `listen` to changes, consider moving the `Provider` call out of the method
+  /// and then using the `value` itself in the [createBloc].
+  /// Don't forget to also reference the `value` in the [inputs].
```
tp: Should we add a line like:

> If the value of the `Provider` is a `ValueListenable`, you don't have to listen through the provider, but can subscribe to the value itself.

The above is a little rough, but the point would be that people don't start moving stuff outside of the `WithBloc.createBloc`when it's not strictly necessary.

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -22,6 +22,13 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
   ///
   /// This function might be called multiple times in the lifetime of this widget.
   /// Whenever the [inputs] change, a new BLoC will be created by calling this function.
+  ///
+  /// For creating the initial bloc, this function will be called in the `initState`.
+  /// That means that you can listen to any `Provider`s or `MediaQuery`.
```
tp:
```suggestion
  /// That means that you can not listen to any `Provider`s or `MediaQuery`.
```

packages/with_bloc/lib/src/with_bloc.dart
HenriBeck: I think the start of the sentence makes this clear with:
> If you need to `listen` to changes

tp: But what I want to suggest to the reader, is not listening on the widget-level and using `inputs` to recreate the bloc, but rather take the full `ValueNotifier<T>` in the constructor of the bloc and subscribe inside the bloc itself. (So the bloc doesn't get recreated all the time.)

PR #25, opened 2019-12-21
Add analyze options and script

HenriBeck:
#23

### DIFF
--- a/packages/state_queue/analysis_options.yaml
+++ b/packages/state_queue/analysis_options.yaml
@@ -1,0 +1,146 @@
+include: package:pedantic/analysis_options.yaml
+
+linter:
+  rules: # http://dart-lang.github.io/linter/lints/options/options.html
+    - always_put_control_body_on_new_line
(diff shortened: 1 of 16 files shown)


packages/state_queue/analysis_options.yaml
```diff
@@ -0,0 +1,146 @@
...
+
+  exclude:
+    - lib/**/*.g.dart
+    - lib/**/*.pb.dart
+    - lib/**/*.pbenum.dart
+    - lib/**/*.pbgrpc.dart
+    - lib/**/*.pbjson.dart
+    - lib/**/*.pbserver.dart
+    - packages/**
+    - lib/ay_icons_icons.dart
```
tp: Not needed, should we slim these a little?

tp:
Sorry for the delay. Good stuff. LGTM.

packages/state_queue/analysis_options.yaml
HenriBeck: Yes totally makes sense

Issue #24, opened 2019-12-21
[with_bloc] RFC: `onInputsChange`


tp:
I just came across an idea of how to make the task updating a bloc based on parent inputs cleaner in the case where the bloc shouldn't be recreated completely, but just update itself.

The current option was basically to pass a `ValueNotifier` to the bloc, which could then do it's own subscribing and updating. This lost the declarative convenience of `WithBloc` and forced the bloc to accept some "management overhead" distracting from the core flow.

So, what about an `onInputsChange` callback, that is invoked (if present) when the inputs change, instead of recreating the bloc entirely?

I think still creating a full initially by `createBloc` is still preferable over a pattern that relies on the updater also being called for the first render (as you would then have to have some empty/"loading" state in the bloc). (This is also aligned with how we handle our `Provider` creation.)

cc @HenriBeck

HenriBeck:
Sounds good.

Though how can the user than differentiate between recreating the bloc and only updating it? 
So assuming that some input changes might cause an update rather than a recreation, but for others, the user might desire to recreate the bloc.

Or when passing an `onInputsChange` you opt entirely out of the recreation of the bloc?

tp:
> Or when passing an onInputsChange you opt entirely out of the recreation of the bloc?

Exactly, that was the idea.

So `createBloc` would only be called once if `onInputsChange` is provided.

I don't see a way to _only_ have one or the other, except with a pattern like the current `Provider` that creates a "not fully initialized" bloc and then updates it via some setters.

---

An alternative idea that just popped into my head would be something like `createBloc(previousBloc)`, where you could do something like:

```dart
createBloc: (previousBloc) {
  if (previousBloc != null) {
    return previousBloc..updateX(x);
  }

  return Bloc(x: x);
}
```

PR #26, opened 2020-02-09
onInputsChange RFC

HenriBeck:
#24


tp:
@HenriBeck Sorry, don't have proper notifications for our core repos. Will review this tomorrow.

HenriBeck:
@tp ping

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -22,12 +23,19 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
   ///
   /// This function might be called multiple times in the lifetime of this widget.
   /// Whenever the [inputs] change, a new BLoC will be created by calling this function.
+  ///
+  /// For creating the initial BLoC, this function will be called in the `initState`.
+  /// That means that you can not listen to any `Provider`s or `MediaQuery`.
+  ///
+  /// If you need to `listen` to changes, consider moving the `Provider` call out of the method
+  /// and then using the `value` itself in the [createBloc].
+  /// Don't forget to also reference the `value` in the [inputs].
```
tp:
```suggestion
  /// Don't forget to also reference the `value` in the [inputs] so the bloc would be rebuilt if it changes.
```

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -48,8 +56,45 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
 
   /// The parameters the BLoC depends upon.
   ///
+  /// If [onInputsChange] is provided, the [onInputsChange] callback will be called when these inputs change.
+  ///
+  /// Otherwise we will create a new BLoC when the [inputs] change and dispose the old BLoC.
   /// If these change, the BLoC will be recreated and the old BloC will be disposed.
```
tp:
```suggestion
```

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -22,12 +23,19 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
   ///
   /// This function might be called multiple times in the lifetime of this widget.
   /// Whenever the [inputs] change, a new BLoC will be created by calling this function.
```
tp:
```suggestion
  /// Whenever the [inputs] change, a new BLoC will be created by calling this function,
  /// unless an [onInputsChange] callback was provided, in which case the latter is used to update the existing bloc.
```

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -48,8 +56,45 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
...
   /// If these change, the BLoC will be recreated and the old BloC will be disposed.
-  final List<dynamic> inputs;
+  final List inputs;
+
+  /// This callback will be called when ever the [inputs] change.
+  /// This will receive the current [BlocType], the previous [inputs] and the new [inputs].
+  ///
+  /// Can be `null`.
+  ///
+  /// Use this callback to update (call some method) instead of the default behaviour of recreating the BLoC.
```
tp:
```suggestion
  /// Use this callback to update the BLoC (e.g. call some method) instead of the default behavior of creating a new BLoC instance.
```

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -48,8 +56,45 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
...
+  ///
+  /// The return value should be a [bool], this will decide if we recreate the BLoC based on the inputs change.
```
tp:
```suggestion
  /// The returned [bool] signals whether the bloc has been updated successfully with the new inputs,
  /// or whether a new instance should be created from scratch.
```

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -48,8 +56,45 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
...
+  /// Return `true` for recreating the BLoC and `false` for not recreating it.
+  ///
+  /// This allows you to handle different input changes differently, e.g.:
+  /// {@tool sample}
+  ///
+  /// ```dart
+  /// WithBloc(
+  ///   createBloc: () => Bloc(),
+  ///   inputs: [locale, isVip],
+  ///   onInputsChange: (bloc, {previousInputs, newInputs}) {
```
tp: @HenriBeck I think I would have inverted this. Return `true ` to signal that the callback has done it's job successfully.

packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -63,21 +108,37 @@ class WithBlocState<BlocType extends ValueNotifier<StateType>, StateType>
...
+          bloc,
+          previousInputs: oldWidget.inputs,
+          newInputs: widget.inputs,
+        );
+      }
+
+      if (recreateBloc) {
+        _disposeBloc();
+
+        /// Recreate the BLoC
```
tp:
```suggestion
        /// Create a new BLoC instance
```

packages/with_bloc/lib/src/with_bloc.dart
HenriBeck: Yeah, sounds also good. Then it's also in sync with how the `Notifications` in Flutter work, e.g. returning true to cancel the bubbling

PR #27, opened 2020-03-10
Sign in with 

tp:


tp:
@HenriBeck Stuck here, as the Swift method channel handler presumably never gets executed. Not sure if I missed to connect some piece here, but it just stalls (also no error on the Flutter side, so I would suppose it was successful in calling something…)

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -0,0 +1,68 @@
...
+            let appleIDProvider = ASAuthorizationAppleIDProvider()
+            let request = appleIDProvider.createRequest()
+            request.requestedScopes = [.fullName, .email]
+            
+            let passwordProvider = ASAuthorizationPasswordProvider()
+            let passwordRequest = passwordProvider.createRequest()
+            
+            let authorizationController = ASAuthorizationController(authorizationRequests: [passwordRequest, request])
+            authorizationController.delegate = self
+            authorizationController.performRequests()
```
HenriBeck: @tp you will need to call here `result` to complete the method, otherwise, the call on the flutter side will never complete

HenriBeck: We also probably want to return some sort of status or error so the UI can react to it

tp:
@HenriBeck Feel free to have a look now at the code.

Open things:
- Add tests, though without a properly set up device we can only do some unit test on the parsing, which is not ideal as the most error prone part is using the same strings between Swift and Dart
- Update example to actually render the SiwA button (also need to check if this would ever be function without the proper certificate setup)
- Allow handler to be passed to the button, so it doesn't just do dummy prints

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -0,0 +1,124 @@
...
+  final String userIdentifier;
+
+  /// Can be `null`
+  final String givenName;
+
+  /// Can be `null`
+  final String familyName;
+
+  /// Can be `null`
+  final String email;
```
HenriBeck: In what circumstances could this be nil? In case the user is already registered?

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -0,0 +1,124 @@
...
+        userIdentifier: response['userIdentifier'],
+        givenName: response['givenName'],
+        familyName: response['familyName'],
+        email: response['email'],
+      );
+
+    case 'password':
+      return AuthorizationPassword(
+        username: response['username'],
+        password: response['password'],
```
HenriBeck: Shouldn't this complain when analyzing as this is implicitly casting `dynamic` into `String`?

packages/sign_in_with_apple/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
...
+# following page: https://dart.dev/tools/pub/pubspec
+
+# The following section is specific to Flutter.
+flutter:
+  # This section identifies this Flutter project as a plugin project.
+  # The androidPackage and pluginClass identifiers should not ordinarily
+  # be modified. They are used by the tooling to maintain consistency when
+  # adding or updating assets for this project.
+  plugin:
+    androidPackage: com.example.sign_in_with_apple
```
HenriBeck: It would be nice if this could also be changed to `de.aboutyou.mobile.app`, but not.a big deal

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
tp: It works, but maybe the analyzer is not as string. I was just assuming it does an automatic type check / conversion.

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
tp: You only get e-mail and name once. If you would then show the sheet again with the app already being associated with the user's Apple ID, one only gets the user identifier back.

tp:
Also we should find a nice way to offer an overall `isSignInWithApple` available, so people can conditionally render the button.

Doesn't seem that simple (for the usage site), as we can only get the device's OS version `async` (e.g. using `device_info` package).

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
HenriBeck: Should we then maybe model this here just to make it clear?

HenriBeck:
> Also we should find a nice way to offer an overall `isSignInWithApple` available, so people can conditionally render the button.
> 
> Doesn't seem that simple (for the usage site), as we can only get the device's OS version `async` (e.g. using `device_info` package).

This should work:
```
/// Returns a future
final isAvailableFuture = AppleSignIn.isAvailable();

class IfAppleSignInIsAvailable extends StatelessWidget {
  const IfAppleSignInIsAvailable({
    Key key,
    @required this.child,
    @required this.fallback,
  })  : assert(child != null),
        super(key: key);

  /// The [Widget] which should be rendered if Apple Sign In is available.
  final Widget child;

  /// Can be `null`
  final Widget fallback;

  @override
  Widget build(BuildContext context) {
    return FutureBuilder<bool>(
      future: isAvailableFuture,
      builder: (context, snapshot) {
        if (snapshot.hasData && snapshot.data) {
          return child;
        }

        return fallback;
      },
    );
  }
}
```

This will call the method channel when the app starts, and in most cases, the future should complete before the sign-in is rendered.

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
tp: You mean 2 Apple ID credentials classes, with and without username/password/email?

HenriBeck: exactly, I think that would make it easier on our app side then to show the intent / possibilities

tp:
The renderer widget incl. a check for availability via the method channel sounds good. That way we might also be able to check whether the user's Apple ID is set up correctly so it would actually show the Apple ID screen.

Would do this as a follow up though, to not grow this further. For testing we can just to a simple `isIOS` and `startsWith('13.')` check.

HenriBeck:
> The renderer widget incl. a check for availability via the method channel sounds good. That way we might also be able to check whether the user's Apple ID is set up correctly so it would actually show the Apple ID screen.

Sounds good, should I have a look again?

tp:
> Sounds good, should I have a look again?

Yeah, please do.

Fine from my side for a v1. The only know issue is the error handling part, where we can't discern between the password request going wrong, or the user actively cancelling.

This means that when the user actively cancels, we would show the sheet a second time (as we assume the password request went wrong).

Can't think of a good workaround here, as we don't get to cast to the "inner error class" (which would have more details).

Maybe a decent hack would be to check whether the rejection happened within a second, and if so think it was an automatic rejection – above 1s it would then be considered user interaction.

Will also create a follow up for that approach.

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
tp: Follow up created.

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -0,0 +1,199 @@
...
+        var requests: [ASAuthorizationRequest] = [request]
+
+        if currentAction == .attemptWithPasswords {
+            let passwordProvider = ASAuthorizationPasswordProvider()
+            let passwordRequest = passwordProvider.createRequest()
+
+            requests.append(passwordRequest)
+        }
+
+        let authorizationController = ASAuthorizationController(authorizationRequests: requests)
```
HenriBeck: So if one of the requests fails, the `didCompleteWithError` will be called?

And if we have a request with a password, we then try to request without it?

packages/sign_in_with_apple/ios/sign_in_with_apple.podspec
```diff
@@ -0,0 +1,23 @@
...
+Pod::Spec.new do |s|
+  s.name             = 'sign_in_with_apple'
+  s.version          = '0.0.1'
+  s.summary          = 'A new flutter plugin project.'
+  s.description      = <<-DESC
+A new flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
+  s.license          = { :file => '../LICENSE' }
+  s.author           = { 'Your Company' => 'email@example.com' }
```
HenriBeck:
```suggestion
  s.author           = { 'Timm Preetz' => 'timm.preetz@aboutyou.com' }
```

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -0,0 +1,45 @@
...
+    show
+        AuthorizationCredential,
+        AuthorizationCredentialAppleID,
+        AuthorizationCredentialPassword;
+export './credential_state.dart' show CredentialState;
+export './sign_in_with_apple_button/sign_in_with_apple_button.dart'
+    show SignInWithAppleButton, SignInWithAppleButtonStyle, IconAlignment;
+
+class SignInWithApple {
+  static const MethodChannel _channel =
```
HenriBeck:
```suggestion
  static const MethodChannel channel =
```

Might be nice to expose the channel so we can mock it in tests and assure that we call a method on that channel

packages/sign_in_with_apple/ios/sign_in_with_apple.podspec
```diff
@@ -0,0 +1,23 @@
...
+# Run `pod lib lint sign_in_with_apple.podspec' to validate before publishing.
+#
+Pod::Spec.new do |s|
+  s.name             = 'sign_in_with_apple'
+  s.version          = '0.0.1'
+  s.summary          = 'A new flutter plugin project.'
+  s.description      = <<-DESC
+A new flutter plugin project.
+                       DESC
+  s.homepage         = 'http://example.com'
```
HenriBeck:
```suggestion
  s.homepage         = 'https://github.com/aboutyou/dart_packages/tree/master/packages/sign_in_with_apple'
```

packages/sign_in_with_apple/ios/sign_in_with_apple.podspec
```diff
@@ -0,0 +1,23 @@
+#
+# To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html.
+# Run `pod lib lint sign_in_with_apple.podspec' to validate before publishing.
+#
+Pod::Spec.new do |s|
+  s.name             = 'sign_in_with_apple'
+  s.version          = '0.0.1'
+  s.summary          = 'A new flutter plugin project.'
+  s.description      = <<-DESC
+A new flutter plugin project.
```
HenriBeck:
```suggestion
Flutter bridge to initiate Sign in with Apple (currently iOS only). Includes support for keychain entries as well as sign in with an Apple ID.
```

packages/sign_in_with_apple/ios/sign_in_with_apple.podspec
```diff
@@ -0,0 +1,23 @@
+#
+# To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html.
+# Run `pod lib lint sign_in_with_apple.podspec' to validate before publishing.
+#
+Pod::Spec.new do |s|
+  s.name             = 'sign_in_with_apple'
+  s.version          = '0.0.1'
+  s.summary          = 'A new flutter plugin project.'
```
HenriBeck:
```suggestion
  s.summary          = 'Flutter plugin for handling Sign in with Apple'
```

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
tp: Yup, that's exactly how this works.

Tried a bunch of attempts to get the correct error code, so we could fail early if appropriate, but couldn't figure it out, as the detailed error class is just internal.

Also seems broken from the API perspective, as Apple recommends to request both to not create a second account for users who already have one. But since there is no open bug tracker I just have that forums comment to work with (which explained the error, but didn't suggest any workaround / alternative approach).

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
tp: I know that you introduced that pattern to the app, but was there any guidance on this?

Thinking about this, really surprised how both the static channel on the `SignInWithApple` class and then the channel created in tests don't interfere with each other.

HenriBeck: You can technically create multiple channels for the same channel name.

They will just send the data then with the same channel name to the native side. 
By default they use a shared messenger, so you can create two instances of a channel, set a mock handler on the first one, send data on the other one and the first would receive the call.

I think depending on just the channel name is probably worse, as that should be even a bigger implementation detail.

PR #31, opened 2020-03-13
Provide conditionally rendering widget

HenriBeck:


packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -44,4 +46,8 @@ class SignInWithApple {
       ),
     );
   }
+
+  static Future<bool> isAvailable() {
+    return channel.invokeMethod<bool>('isAvailable');
```
tp:
```suggestion
    return _isAvailable ??= channel.invokeMethod<bool>('isAvailable');
```

So we cache this for cases where the button is not always displayed?

Then the signature could even become `FutureOr<bool>` and we an immediate value in most cases.

tp: @HenriBeck Ah, it's cached on the widget (but then still wrapped inside a Future). Does `FutureBuilder` have some private access to directly render with a resolved future? Didn't see any public API for that so I would be surprised…

HenriBeck:
@tp Maybe we need to use a `SynchronousFuture<bool>` in case we have the value cached?

tp:
```
static FutureOr<bool> isAvailable() {
   return _isAvailable ??= await channel.invokeMethod<bool>('isAvailable');
}
```

Might work, no? (with _isAvailable being `bool`.) Otherwise we'd have to use the special logic in the widget, so it would use the bool if it's available.

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -47,7 +48,24 @@ class SignInWithApple {
     );
   }
 
-  static Future<bool> isAvailable() {
-    return _isAvailable ??= channel.invokeMethod<bool>('isAvailable');
+  /// Cached value whether or not sign-in with apple is available
+  ///
+  /// We cache, so we can return a [SynchronousFuture] in case this value has already been loaded
```
tp: @HenriBeck 

Even `SynchronousFuture` is not handled specially by the `FutureBuilder` (https://github.com/flutter/flutter/blob/0b8abb4724/packages/flutter/lib/src/widgets/async.dart#L566)

So I guess the only solution would be to not use the FB if we have a value.

HenriBeck: @tp It doesn't have to handle it. The `SynchronousFuture ` executes the `then` callback synchronously, which should result in the widget not needing a frame to resolve the future.

tp: You're right @HenriBeck. It will call `_subscribe` during `initState`, in which then `then` and hence `setState` are executed synchronously, so the data will be there for the first rendering.

https://github.com/flutter/flutter/blob/0b8abb4724/packages/flutter/lib/src/widgets/async.dart#L661

PR #36, opened 2020-03-29
macOS support

tp:


packages/sign_in_with_apple/example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -0,0 +1,12 @@
...
+//  Generated file. Do not edit.
+//
+
+import FlutterMacOS
+import Foundation
+
+import sign_in_with_apple
+
+func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
+    SignInWithApplePlugin.register(with: registry.registrar(forPlugin: "SignInWithApplePlugin"))
```
tp: In order to get this working on macOS at all, I had to manually edit this file (and then run via Xcode instead of Flutter, as the latter would just overwrite it again).

Furthermore I had to put the Obj-C / header bridge files in place, even though for example `url_launcher` is just plain Swift. Not sure if that works a little differently because it has a `macOS` specific package, or if something in these files still suggest Obj-C…

Flutter for macOS also needs a globally installed cocoapods for now to detect the plugin dependencies, but in this case that still doesn't work (even though a `bundle pod install` does add the dependency to the Xcode project).

cc @HenriBeck

packages/sign_in_with_apple/example/macos/Podfile.lock
```diff
@@ -0,0 +1,16 @@
+PODS:
+  - FlutterMacOS (1.0.0)
+
+DEPENDENCIES:
+  - FlutterMacOS (from `Flutter/ephemeral/.symlinks/flutter/darwin-x64`)
+
+EXTERNAL SOURCES:
```
tp: Ah, while the `sign_in_with_apple` plugin once was here, it now disappeared again. That would explain the above error.

tp: Ah, for macOS we need the new format in pubspec.yaml, while that didn't work for iOS for me…

packages/sign_in_with_apple/example/macos/Flutter/GeneratedPluginRegistrant.swift
tp: Fixed in e57894321cda7732a48a4b27dddcded58dd23849

packages/sign_in_with_apple/example/macos/Gemfile
```diff
@@ -0,0 +1,10 @@
+source "https://rubygems.org"
+
+plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
+eval_gemfile(plugins_path) if File.exist?(plugins_path)
+
+gem "cocoapods"
+
+gem "fastlane"
```
HenriBeck: We only need cocoapods here, right?

packages/sign_in_with_apple/macos/Classes/SignInWithApplePlugin.swift
```diff
@@ -0,0 +1,208 @@
...
+            default:
+                result(FlutterMethodNotImplemented)
+
+                return
+            }
+        } else {
+            result(
+                FlutterError(
+                    code: "ERR",
+                    message: "Unsupported iOS version",
```
HenriBeck:
```suggestion
                    message: "Unsupported macOS version",
```

tp:
@HenriBeck The remaining issue with the example is that one can't execute it locally without an Apple developer account (Apple ID for local development is not enough).

For the iOS plugin I was testing it through our app, but here I wasn't able to get it working yet without messing up the other certificates again (when letting Xcode "fix" it). But at least I am getting an Apple error when running it locally, which points to the capability issue.

packages/sign_in_with_apple/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: sign_in_with_apple
-description: Flutter bridge to initiate Sign in with Apple (currently iOS only). Includes support for keychain entries as well as sign in with an Apple ID.
-version: 1.0.0
+description: Flutter bridge to initiate Sign in with Apple (currently iOS and macOS only). Includes support for keychain entries as well as signing in with an Apple ID.
+version: 1.1.0
 homepage: https://github.com/aboutyou/dart_packages/tree/master/packages/sign_in_with_apple
 repository: https://github.com/aboutyou/dart_packages
```
HenriBeck: @tp I think we should still have a flutter version constraint here

tp: Reverted, as it was the latest stable anyway and the one we tested with.

PR #39, opened 2020-04-10
[state_queue] Use Zones for StateQueue error handling

HenriBeck:


packages/state_queue/lib/src/state_queue.dart
```diff
@@ -76,15 +76,10 @@ abstract class StateQueue<T> extends ValueNotifier<T>
...
-        } catch (e, stack) {
-          if (StateQueue.errorHandler != null) {
-            StateQueue.errorHandler(e, stack);
-          } else {
-            print(
-              'StateQueue: Error occured in `run` stream:\nimplementation: $this\n$e\n$stack',
-            );
-          }
+        } catch (error, stack) {
+          Zone.current.handleUncaughtError(error, stack);
```
HenriBeck, line 82: I think using `Zone` here might be a better general option here, as it allows for composition and is built into dart.

tp: Seems fine, especially since we just divert them to the same handler in the app anyway (which seems like the common case).

We just want to get them out of here, so that future `run`s still work.

tp:
@HenriBeck: + Version bump (breaking change) + CHANGELOG entry please

Issue #34, opened 2020-03-22
[sign_in_with_apple] Allow to request authentication without username/password


tp:
In case Sign in with Apple returns saved keychain credentials (username/password) which are not valid on the backend the app should have the possibility to request login with an Apple ID (and ignore the saved credentials).

Otherwise this would lead to SiwA being stuck/not available to people who have old credentials in their keychain.

By providing this switch in the API, the app can decide whether to proceed with a "password reset" flow, or to create a new account based on the Apple ID.

HenriBeck:
We should also allow specifying the scopes which are used for the apple sign in. Currently, those are hardcoded to both email and the full name, which might not be required for all apps.

HenriBeck:
Closing this because we have split up the APIs

Issue #30, opened 2020-03-11
[sign_in_with_apple] Refine error handling for "password not available case"


tp:
Currently we always treat the first error reported by `AuthKit` as expected, as it happens that when you request password + Apple ID credentials and the user doesn't have any passwords saved, you would just get a generic error (even though the system would only ever return one or the other credentials anyway…).

This leads to the situation that the first manual cancellation, wich is also only reported as a generic error (status 1000), causes the sheet to be shown a second time.

To workaround this, we could only treat errors within the first second as automatic rejections, and everything else and manual cancellations which should be respected.

Reference: https://github.com/aboutyou/dart_packages/pull/27#issuecomment-597757215

HenriBeck:
I think we can also provide better errors as there is a mapping on the native side: https://developer.apple.com/documentation/authenticationservices/asauthorizationerror/code

HenriBeck:
@tp After looking more at the whole error handling of the authorization, maybe it also makes sense to only support a single request per authorization on the Dart side.

My reasoning behind this: I didn't find a way of providing both Password and an Apple ID request where I wouldn't run into an error in case no password is stored.

tp:
> My reasoning behind this: I didn't find a way of providing both Password and an Apple ID request where I wouldn't run into an error in case no password is stored.

We can definitely simplify this on the native side (to only allow requesting 1 at a time), so get more logic on the Dart side. I would still allow the consumer to request a list of authentications with a priority/order, so that the usage is very simple.

HenriBeck:
@tp With the improved API, I think we can close this.

tp:
@HenriBeck I still think it has 1 ugliness by throwing for a missing Keychain entry instead of returning `null`. But will verify and attempt a fix for that later. Consumers should definitely execute each login attempt in an individual try/catch.

HenriBeck:
@tp I think throwing an error is more appropriate here, to be honest. I would consider this a success flow, and as you said, the user should wrap each call in a `try/catch`.

So I think it's easier to throw an error and not have to handle the `null` case in the success flow (even with NNDB coming up)

PR #42, opened 2020-04-10
[sign-in-with-apple] Allow specifying requests and scopes

HenriBeck:
#30 #34


packages/sign_in_with_apple/lib/src/authorization_request.dart
```diff
@@ -0,0 +1,80 @@
...
+
+/// A base class which describes an authorization request that we will make on the native side.
+///
+/// Apple Docs: https://developer.apple.com/documentation/authenticationservices/asauthorizationrequest
+///
+/// We currently only support the following authorization requests:
+/// - [AppleIDAuthorizationRequest]
+/// - [PasswordAuthorizationRequest]
+@immutable
+abstract class AuthorizationRequest {
```
HenriBeck, line 11: @tp Please have a look here, especially on the documentation side

packages/sign_in_with_apple/lib/src/exceptions.dart
```diff
@@ -0,0 +1,85 @@
+import 'package:flutter/services.dart';
+import 'package:meta/meta.dart';
+
+@immutable
+abstract class SignInWithAppleException implements Exception {
```
HenriBeck, line 5: @tp I also started out creating some proper exception types which should make the error handling more pleasant for the users and easier to discover which errors could occur.

I'm not sure how fine-grained these need to be though.

packages/sign_in_with_apple/lib/src/authorization_request.dart
tp: Gladly, will do it first thing in the morning!

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -3,10 +3,14 @@ import Flutter
 import UIKit
 
 public class SwiftSignInWithApplePlugin: NSObject, FlutterPlugin {
-    var _lastAYSignInWithAppleAuthorizationController: Any? // will be `AYSignInWithAppleAuthorizationController` in practice, but we can't scope the variable to iOS13+
+    // will be `AYSignInWithAppleAuthorizationController` in practice, but we can't scope the variable to iOS13+
```
tp, line 6:
```suggestion
    // will be `SignInWithAppleAuthorizationController` in practice, but we can't scope the variable to iOS13+
```

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -95,13 +115,11 @@ public class SwiftSignInWithApplePlugin: NSObject, FlutterPlugin {
...
             default:
                 result(FlutterMethodNotImplemented)
-
-                return
             }
         } else {
             result(
                 FlutterError(
-                    code: "ERR",
+                    code: "not-supported",
```
tp, line 122: Should we group all error codes in a single location as enum constants? Saw that in another package and makes it great to check what incoming data to expect in the Dart side

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -111,99 +129,161 @@ public class SwiftSignInWithApplePlugin: NSObject, FlutterPlugin {
...
+                let appleIDProvider = ASAuthorizationAppleIDProvider()
+                let appleIDRequest = appleIDProvider.createRequest()
+                
+                if let scopes = requestMap["scopes"] as? [String] {
+                    appleIDRequest.requestedScopes = []
+                    
+                    for scope in scopes {
+                        switch scope {
+                        case "email":
+                            appleIDRequest.requestedScopes?.append(.email)
```
tp, line 170: Force unwrap as we just initialized it above? Or else we could build out own list, and set the final value to the request?

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -24,29 +33,60 @@ class SignInWithApple {
...
   }
 
+  /// Request the credentials state for the user.
+  ///
+  /// This methods either completes with a [CredentialState] or throws an [SignInWithAppleException].
+  /// In case there was an error while getting the credentials state, this throws a [SignInWithAppleCredentialsException].
+  ///
+  /// Apple Docs: https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidprovider/3175423-getcredentialstate
   static Future<CredentialState> getCredentialState(
     String userIdentifier,
```
tp, line 76: Should mention how the user identifier comes from an earlier Apple ID credentials request response

packages/sign_in_with_apple/lib/src/authorization_request.dart
```diff
@@ -0,0 +1,80 @@
...
+  final List<AppleIDAuthorizationScopes> scopes;
+
+  @override
+  String toString() => 'AppleIDAuthorizationRequest(scopes: $scopes)';
+
+  @override
+  Map<String, dynamic> toJson() {
+    return <String, dynamic>{
+      'type': 'appleid',
+      'scopes': scopes
```
tp, line 45: List literal with ifs?

packages/sign_in_with_apple/macos/Classes/SignInWithApplePlugin.swift
```diff
@@ -100,7 +100,7 @@ public class SignInWithApplePlugin: NSObject, FlutterPlugin {
         } else {
             result(
```
tp, line 101: For the macOS plugin I thought about reusing the same source code file, potentially with some conditional compilation if needed

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -111,99 +129,161 @@ public class SwiftSignInWithApplePlugin: NSObject, FlutterPlugin {
...
+                    code: "unknown-credentials",
                     message: "Unexpected credentials: \(authorization.credential)",
                     details: nil
                 )
             )
         }
     }
 
-    public func authorizationController(controller _: ASAuthorizationController, didCompleteWithError error: Error) {
-        if currentAction == .attemptWithPasswords {
```
tp, line 188: So is the Flutter side now expected to make 2 request to first query the keychain, and if that fails try the Apple ID? I think we might just keep trying one provider after another, else everyone would be writing the same code in their app (first try keychain, then Apple ID).

But giving the user control over the methods and omit some method entirely is great 👍

packages/sign_in_with_apple/macos/Classes/SignInWithApplePlugin.swift
HenriBeck: I was thinking the same. The biggest issue I found is that you can't include external files in a `Podspec`.

Maybe we just need a separate `Podspec` which is also `Flutter` independent?

What I was also thinking is about having two classes for the Plugin. One that is used when signing in with apple is not available, and one that is used when it's possible.

That would improve the usages of our `#available` as those are spread around.

packages/sign_in_with_apple/lib/src/authorization_request.dart
HenriBeck: Looks better 👍

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
HenriBeck: Building our own list was a bit more explicit and I removed it.

While we know that this shouldn't be `nil` here, I don't see a big benefit here by force unwrapping, which could lead to a full crash.

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
HenriBeck: Yeah sounds good 👍

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
HenriBeck: Do you know if the success callback can be called additionally to the error callback? Is that a valid flow?

I find the multiple requests a bit weird, but I do want to do some testing later today here anyways.

packages/sign_in_with_apple/macos/Classes/SignInWithApplePlugin.swift
HenriBeck: I found the following docs on the sharing code issue:
https://github.com/flutter/flutter/issues/52285
https://github.com/flutter/plugins/pull/2618/files?file-filters%5B%5D=.dart&file-filters%5B%5D=.entitlements&file-filters%5B%5D=.h&file-filters%5B%5D=.m&file-filters%5B%5D=.md&file-filters%5B%5D=.plist&file-filters%5B%5D=.png&file-filters%5B%5D=.podspec&file-filters%5B%5D=.swift&file-filters%5B%5D=.xcconfig&file-filters%5B%5D=.xib&file-filters%5B%5D=.yaml&file-filters%5B%5D=dotfile

packages/sign_in_with_apple/lib/src/exceptions.dart
tp: Probably  a little too specific in practice, but also does no harm as everything is under the `SignInWithAppleException` umbrella.

I think the only thing that is relevant is to clearly separate "something is broken" (like no SiwA support due to now having 2FA set up) from "the user cancelled".

packages/sign_in_with_apple/macos/Classes/SignInWithApplePlugin.swift
tp: Nice. Also the iOS version still has the bridging header while macOS does not. Don't recall if that was required, or just how the template was generated (as there was quite some back and forth when trying to get the initial implementation building).

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
tp: > Do you know if the success callback can be called additionally to the error callback? Is that a valid flow?

I don't know that's is prohibited, but it would for sure be unexpected to me.

HenriBeck: While I was looking into this, I didn't find any good reason for actually providing two auth requests.
Once I provided a password request, I ended up always having an unknown error. Only if I removed the password request, I got the apple sign in again.
So I guess this API change should be fine and the end-users could decide which request they want to make when.

packages/sign_in_with_apple/macos/Classes/SignInWithApplePlugin.swift
HenriBeck: I will look into this in a future PR

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
HenriBeck: Will create a follow up PR to review the general docs of this package

packages/sign_in_with_apple/lib/src/exceptions.dart
HenriBeck: I think we can further improve the docs here in a follow up PR

HenriBeck:
@tp If it's fine with you, I would port over the changes here to the macOS platform in a follow-up PR where I look into code sharing

packages/sign_in_with_apple/example/lib/main.dart
```diff
@@ -24,7 +24,16 @@ class _MyAppState extends State<MyApp> {
           child: Center(
             child: SignInWithAppleButton(
               onPressed: () async {
-                final credentials = await SignInWithApple.requestCredentials();
+                final credentials = await SignInWithApple.requestCredentials(
```
tp, line 27: Wondering whether we should make this simpler the API, so the Apple ID scopes can be directly set.

Don't see this changing a lot over time / gaining much configurability.

I think we could even handle the Keychain fallback with a `preferExistingKeychainEntry` parameter (diverging a little from the Apple-provided API). I think that makes it a lot simpler, is and actually nicer for the specific SiwA case (don't think people would try this library to _only_ get Keychain credentials).

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
```diff
@@ -2,11 +2,96 @@ import AuthenticationServices
...
+    case notSupported(String)
+    
+    case missingArguments(FlutterMethodCall)
+    
+    case missingArgument(FlutterMethodCall, String)
+    
+    @available(iOS 13.0, *)
+    case credentialsError(String)
+    @available(iOS 13.0, *)
+    case unexpectedCredentialsState(ASAuthorizationAppleIDProvider.CredentialState)
```
tp, line 15:
```suggestion
    case unexpectedCredentialState(ASAuthorizationAppleIDProvider.CredentialState)
```

packages/sign_in_with_apple/ios/Classes/SwiftSignInWithApplePlugin.swift
tp: Certainly don't want to crash, but I think using `?` (which might to nothing) in cases where actions are expected could lead to hard-to-discover bugs.

Here it's so close and direct that we know it'll work, but I would be wary if the places are further apart.

packages/sign_in_with_apple/lib/src/exceptions.dart
```diff
@@ -0,0 +1,141 @@
...
+  ) {
+    switch (exception.code) {
+      case 'not-supported':
+        return SignInWithAppleNotSupportedException(
+          message: exception.message,
+        );
+
+      /// Exceptions which indicate an [SignInWithAppleAuthorizationError]
+      case 'authorization-error/unknown':
+        return SignInWithAppleAuthorizationError(
```
tp, line 23: These should all be `Exception`s on the Dart side IMHO (https://groups.google.com/a/dartlang.org/forum/#!msg/misc/lx9CXiV3o30/s5l_PwpHUGAJ etc.), as they are expected to be caught and the programmer couldn't have prevented them.

HenriBeck: Jup, renamed 👍

packages/sign_in_with_apple/example/lib/main.dart
HenriBeck: Yeah, let's discuss it in #45

PR #43, opened 2020-04-11
Android support

tp:


packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        val url: String? = call.argument("url")
+
+        if (url == null) {
+          result.error("MISSING_ARG", "Missing 'url' argument", call.arguments)
+          return
+        }
+
+        SignInWithApplePlugin.lastAuthorizationRequestResult = result
```
HenriBeck, line 75: @tp I think we should check if we already have an existing `lastAuthorizationRequestResult` here and in case, we should complete that one with an error

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -0,0 +1,131 @@
...
+ * DO NOT rename this or it's package name as it's configured in the consumer's `AndroidManifest.xml`
+ *
+ */
+public class SignInWithAppleCallback: Activity {
+  constructor() : super()
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    SignInWithApplePlugin.lastAuthorizationRequestResult!!.success(intent?.data?.toString())
```
HenriBeck, line 124: I think it would be safer to check here that this is not `null` just so we don't crash here. And in case it is `null`, we should just log some error message

tp:
* I already simplified the external API here quickly (can still further refine this of course).
* Interestingly we don't get the `userIdentifier` and `identityToken` from the web callback. Will look into how this will work on the server then (or if it's also not needed for the Apple platform case).

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
tp: Makes sense, as this can happen when the Custom Tab gets closes by the user.

Will look if this can be detected and resolve the pending task then as well, as that'd be preferable.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
tp: Crashing is fine I think (didn't bring down the app, ran into this a few times while the CCT wasn't getting closed and hence firing the callback multiple times), and then there is at least a chance it will be discovered / reported.

Just logging doesn't seem enough / wouldn't notify anyone.

But we can still check the values properly and throw a more descriptive exception.

tp:
Created #46 as a follow up. Would only be an addition (a nice one though :))

tp:
@HenriBeck Integrated this locally with the app, and retrieving the SiwA `code` worked every time :)

So I think for the client this is done.

HenriBeck:
> @HenriBeck Integrated this locally with the app, and retrieving the SiwA `code` worked every time :)
> 
> So I think for the client this is done.

Nice, I will do the review later tonight.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -0,0 +1,148 @@
...
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+
+    val lastAuthorizationRequestResult = SignInWithApplePlugin.lastAuthorizationRequestResult
+    if (lastAuthorizationRequestResult != null) {
+      lastAuthorizationRequestResult.success(intent?.data?.toString())
+      SignInWithApplePlugin.lastAuthorizationRequestResult = null
+    } else {
+      SignInWithApplePlugin.triggerMainActivityToHideChromeCustomTab = null
```
HenriBeck, line 133: Why are we setting `triggerMainActivityToHideChromeCustomTab ` here to `null`?

HenriBeck: While in practice they could only be `null` together, this seems still out of place

tp: Just as a clean up before we return. Most certainly it's already `null` as this point (with the current code structure at leat)

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -21,35 +32,137 @@ export './src/widgets/sign_in_with_apple_button.dart'
...
+      );
+    }
+
+    if (Platform.isAndroid) {
+      return _signInWithAppleAndroid(
+        scopes: scopes,
+        webAuthenticationOptions: webAuthenticationOptions,
+      );
+    }
+
```
HenriBeck, line 84: Should we check here that we are on iOS or macOS? and if not we should just throw an `Unsupported` exception

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -0,0 +1,107 @@
...
+    throw Exception('Unsupported result type ${response['type']}');
+  }
+}
+
+AuthorizationCredentialAppleID parseAuthorizationCredentialAppleIDFromDeeplink(
+  Uri deeplink,
+) {
+  final user = deeplink.queryParameters.containsKey('user')
+      ? json.decode(deeplink.queryParameters['user']) as Map<String, dynamic>
+      : null;
```
HenriBeck, line 97: TODO: Throw more descriptive errors? Right now this would crash in case the JSON is not formatted properly or the deeplink doesn't contain the `user` param

HenriBeck: nvm

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -0,0 +1,107 @@
...
+  final String givenName;
+
+  /// Can be `null`, will only be returned on the first authorization
+  final String familyName;
+
+  /// Can be `null`, will only be returned on the first authorization
+  final String email;
+
+  /// Can be `null`
+  final String identityToken;
```
HenriBeck, line 31: TODO: document in #44 when this can be `null` and when not

packages/sign_in_with_apple/lib/sign_in_with_apple.dart
```diff
@@ -21,35 +32,137 @@ export './src/widgets/sign_in_with_apple_button.dart'
...
 class SignInWithApple {
   @visibleForTesting
-  static const channel =
-      MethodChannel('de.aboutyou.mobile.app.sign_in_with_apple');
+  static const channel = MethodChannel(
+    'com.aboutyou.dart_packages.sign_in_with_apple',
+  );
+
+  /// Returns the credentials stored in the Keychain for the app-associated websites
+  static Future<AuthorizationCredentialPassword> getKeychainCredential() async {
```
HenriBeck, line 40: I guess with this change we would consider this a breaking change.

Let's release v2 maybe after we fixed the feedback?

PR #47, opened 2020-04-17
Share Swift code for iOS and macOS code

HenriBeck:


HenriBeck:
@tp The code is much nicer with the symlinks now. The platform plugins now only need to decide whether or not `Sign in with Apple` is available, and then register the appropriate plugin.

And the platform-specific checks are limited to the import for the flutter SDK

packages/sign_in_with_apple/darwin/Classes/SignInWithAppleUnavailablePlugin.swift
```diff
@@ -0,0 +1,32 @@
...
+#if os(OSX)
+import FlutterMacOS
+#elseif os(iOS)
+import Flutter
+#endif
+
+public class SignInWithAppleUnavailablePlugin: NSObject, FlutterPlugin {
+    // This plugin should not be registered with directly
+    //
+    // This is merely a cross-platform plugin to handle the case Sign in with Apple is available
```
tp, line 12:
```suggestion
    // This is merely a cross-platform plugin to handle the case Sign in with Apple is not available
```

tp:
Very nice indeed. Also good riddance to the `currentAction` workaround.

PR #50, opened 2020-04-20
SiwA Builder widget

tp:
Use simple stateful widget / builder pattern instead of static field on widget to get SiwA availability

@HenriBeck Not sure how the inherited widget would help here with the very first rendering. Of course it could help with subsequent "first" renderings (when the child widget is added again to a tree).


packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_builder.dart
```diff
@@ -0,0 +1,55 @@
...
+
+/// A widget for conditionally rendering UI based on whether or not Sign in with Apple is available
+class SignInWithAppleBuilder extends StatefulWidget {
+  SignInWithAppleBuilder({
+    Key key,
+    @required this.builder,
+    this.fallbackBuilder = _fallbackBuilder,
+  })  : assert(builder != null),
+        assert(fallbackBuilder != null),
+        // isAvailableFuture = _isAvailable(),
```
HenriBeck, line 16:
```suggestion
```

PR #52, opened 2020-04-20
Use Roboto for golden file test

tp:


packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
```diff
@@ -79,6 +79,9 @@ class SignInWithAppleButton extends StatelessWidget {
                   style: TextStyle(
                     fontSize: height * 0.43 /* per Apple's guidelines */,
                     color: contrastColor,
+                    // defaults styles aligned with https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/cupertino/text_theme.dart
+                    fontFamily: '.SF Pro Text',
```
HenriBeck, line 83: So we removed the SF font and added the roboto font, but then still use the SF font here?

tp: Yeah, it's only that we should not include the font file from Apple due to unclear licensing. It should be fine for testing on macOS, but not cross-platform. So we only "inject" Robot as the SF font for testing.

During normal rendering (on iOS) we want the SF font by default. (Might need to make this configurable though, if it looks out of place in apps that use a different font than the platform default one).

HenriBeck: And what about Android? Do we then also fallback to Roboto?

PR #53, opened 2020-04-20
Run CI on macOS as well and use test matrix

tp:


.github/workflows/dart.yml
```diff
@@ -2,16 +2,24 @@ name: Dart CI
...
+env:
+  DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
 
+jobs:
+  test:
+    name: Test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, macos-10.15]
```
HenriBeck, line 14: I guess in the future we would also want to test across certain stable flutter versions, but that can also be done in a follow-up PR

.github/workflows/dart.yml
```diff
@@ -2,16 +2,24 @@ name: Dart CI
 
 on: [push]
 
-jobs:
-  build:
-    runs-on: ubuntu-latest
-    container:
-      image: cirrusci/flutter:v1.12.13-hotfix.8
+env:
+  DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
```
HenriBeck, line 6: Do we have any control over the XCode version?

tp: Yes:

- https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md
- https://www.jessesquires.com/blog/2020/01/06/selecting-an-xcode-version-on-github-ci/

PR #54, opened 2020-04-20
First docs improvement

tp:
@HenriBeck Reopening this so you can take the lead in reviewing.

The cases where I changed existing comments, it's mostly about the different style (Flutter never uses "we", uses 3rd person to describe the codes behavior, etc. https://dart.dev/guides/language/effective-dart/documentation#doc-comments).

Then I added more comments to some places and added linting so we see what's still missing.

Also the entry point file is now cleaned up, and all internal imports to the `SiwA` plugin static class are now explicitly named, so we ensure that we only import "private" helpers there and don't forget to publicly export something.

Tomorrow I will follow up with another PR for the remaining code, so we can discuss and finish up this one here without having to take new additions into account.


packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -0,0 +1,215 @@
...
+    show
+        parseAuthorizationCredentialAppleID,
+        parseAuthorizationCredentialAppleIDFromDeeplink,
+        parseAuthorizationCredentialPassword;
+import 'package:sign_in_with_apple/src/credential_state.dart'
+    show parseCredentialState;
+
+/// Wrapper class providing the methods to interact with Sign in with Apple.
+// ignore: avoid_classes_with_only_static_members
+class SignInWithApple {
```
tp, line 21: @HenriBeck Thought about dropping the wrapping class entirely, but since it has 4 exported methods, that also seems messy for the consumer (as each function then should be more descriptive. A plain `isAvailable` would be confusing…)

HenriBeck: True, if we would drop the class, we would need to make the function names significantly longer

packages/sign_in_with_apple/README.md
```diff
@@ -30,9 +31,9 @@ SignInWithAppleButton(
 
 ## Flow
```
HenriBeck, line 32: We should potentially differentiate between iOS/macOS and Android here, but I think we can also revamp the README in another PR

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -1,34 +1,69 @@
...
     @required this.givenName,
     @required this.familyName,
     @required this.email,
-    @required this.identityToken,
     @required this.authorizationCode,
   }) : assert(authorizationCode != null);
 
+  /// An identifier associated with the authenticated user.
+  ///
+  /// This will be provided upon every sign-in.
```
HenriBeck, line 29: I guess this statement is contradictory to the `Can be null` in the end.

```suggestion
  /// This will always be provided on iOS and macOS systems. On Android, however, this will not be present.
```

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -1,34 +1,69 @@
...
+  /// The users email in case it was requested.
+  /// You will need to provide the [AppleIDAuthorizationScopes.email] scope to the [AppleIDAuthorizationRequest] for requesting this information.
+  ///
+  /// This information will only be provided on the first authorizations.
+  /// Upon further authorizations, you will only get the [userIdentifier],
+  /// meaning you will need to store this data securely on your servers.
+  /// For more information see: https://forums.developer.apple.com/thread/121496
+  ///
   /// Can be `null`
-  final String identityToken;
```
HenriBeck, line 31: @tp What happened to the identity token? Is this not needed anymore?

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -37,19 +72,23 @@ class AuthorizationCredentialAppleID implements AuthorizationCredential {
...
   @override
   String toString() {
-    return 'AuthorizationAppleID($userIdentifier, $givenName, $familyName, $email, identityToken set? ${identityToken != null}, authorizationCode set? ${authorizationCode != null})';
+    return 'AuthorizationAppleID($userIdentifier, $givenName, $familyName, $email, authorizationCode set? ${authorizationCode != null})';
   }
 }
 
-class AuthorizationCredentialPassword implements AuthorizationCredential {
-  AuthorizationCredentialPassword({
+/// Authorization details retrieved from the user's Keychain for the current app's website.
```
HenriBeck, line 79:
```suggestion
/// Authorization details retrieved from the user's Keychain for the current app's website.
///
/// Only available on iOS and macOS.
```

HenriBeck: Ah, it's already on the `getKeychainCredential`

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -0,0 +1,215 @@
...
+  // ignore: public_member_api_docs
+  static const channel = MethodChannel(
+    'com.aboutyou.dart_packages.sign_in_with_apple',
+  );
+
+  /// Returns the credentials stored in the Keychain for the website associated with the current app.
+  ///
+  /// Only available on Apple platforms.
+  ///
+  /// Throws a [SignInWithAppleException] exception when not credentials have been found in the Keychain.
```
HenriBeck, line 32:
```suggestion
  /// Throws a [SignInWithAppleException] exception when no credentials have been found in the Keychain.
```

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -0,0 +1,215 @@
...
+    @required List<AppleIDAuthorizationScopes> scopes,
+
+    /// Optional parameters for web-based authentication flows on non-Apple platforms
+    ///
+    /// This parameter is required on Android.
+    WebAuthenticationOptions webAuthenticationOptions,
+  }) async {
+    assert(scopes != null);
+
+    if (webAuthenticationOptions == null &&
```
HenriBeck, line 88: should we move this check into the `if (Platform.isAndroid) {` potentially?

packages/sign_in_with_apple/lib/src/authorization_credential.dart
tp: At the moment it's not supported. It would be the JWT token received from the web auth flow, but at AY we don't have any use for it (as we do the server-side validation).

Because of that we also have no good way of testing it. I don't know if this would be useful in any case to use instead of the server-side validated authorization code, but for now I thought it's better to drop it.

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
tp: Done

PR #57, opened 2020-04-22
Add Android to supported platforms in package description

tp:


tp:
We can also release this (or a even catchier version) with the next doc update. Just wanted to get this up as I just noticed it.

PR #59, opened 2020-04-25
Add docs for integration from scratch

tp:


packages/sign_in_with_apple/README.md
```diff
@@ -38,6 +38,87 @@ SignInWithAppleButton(
...
+  * Usually the default setting of "Enable as a primary App ID" should suffice here. If you ship multiple apps that should all share the same Apple ID credentials for your users, please consult the Apple documentation on how to best set these up.
+* Click "Continue", and then click "Register" to finish the creation of the App ID
+
+In case you already have an existing App ID that you want to use with Sign in with Apple:
+
+* Open that App ID from the list
+* Check the "Sign in with Apple" capability
+* Click "Save"
+
+If you have change your app's capabilities, you need to fetch the updated provisioning profiles (for example via Xcode) to use the new capabilities.
```
HenriBeck, line 69: Don't you need to recreate the provisioning profiles? Or is re-fetching really enough?

packages/sign_in_with_apple/README.md
```diff
@@ -38,6 +38,87 @@ SignInWithAppleButton(
...
+* Check the box next to "Sign in with Apple", then click "Configure" on the same row
+* Under "Primary App ID" select the App ID of the app you want to use (either the newly created one or an existing one)
+* Click "Save" to leave the detail view
+* Click "Continue" and then click "Register"
+* Now you'll see a one-time-only screen where you must download the key by clicking the "Download" button
+  * Also note the "Key ID" which will be used later when configuring the server
+
+Now everything is set up on Apple's Developer portal and we can start setting up the server.
+
+### Server
```
HenriBeck, line 103: Really great having an example server for the Android Part 👍 

I would have thought that this would be a big point of question we will get

packages/sign_in_with_apple/README.md
tp: I don't know for sure, as I haven't downloaded these yet (as I really did everything from scratch with a blank new account).

I will follow up with iOS and then macOS separately.

tp:
@HenriBeck I would ask you to review this for Android first.

The example server is now also working fine (had some typos in it).

There is one open question of whether one needs to send the app ID (instead of the service ID) as client ID when using the native app version and how we can detect this, but I will look into that with the iOS example in a follow up.

Just have to quickly do some other things, but I think I can have iOS running from scratch tomorrow morning.

packages/sign_in_with_apple/README.md
```diff
@@ -38,33 +38,128 @@ SignInWithAppleButton(
...
+
+The server part is usually integrated into your existing backends, and there are existing packages for most existing programming languages and web frameworks out there.
+
+In order to show how to build a complete example, we set up a example project on [Glitch](https://glitch.com/) which offers simple and free hosting of a HTTPS-enabled web API, which is exactly what's needed here.
```
HenriBeck, line 107: I think we should make clear that this part redirect part is only needed for Android

HenriBeck: I would suggest moving this into the `Android` section under `Additional Server Requirements` potentially.

packages/sign_in_with_apple/README.md
HenriBeck: Just FYI: While Apple doesn't mention it explicitly, the provisioning profiles are bound to the capabilities of an App while creating that profile. So you will need to recreate these profiles.

In case the user uses the `Automatically manage Signing` within XCode, recreating the profile is not needed as XCode will do it for him.

packages/sign_in_with_apple/README.md
tp: The 2. callback (that the app client example does by itself) is required though, so I think it makes sense to show how it should work. I would think that most people would just put similar code into their own backend and not bother setting up the examples completely.

In the follow up I'll first make it clear that the example server here does and why. We can than still move it (either to another subsection or to server-specific README) [same applies to the overall docs here, maybe we want to move more of the setup to the examples, but I thought at some point it might be handy to be able to compare "here's what the example does" and "here's what you should do"].

PR #61, opened 2020-04-27
iOS integration docs

tp:


tp:
Should be straightforward (with the Xcode-managed certificates), but currently I can't run the example on iOS (it always stays stuck in a white screen).

packages/sign_in_with_apple/README.md
```diff
@@ -162,4 +162,19 @@ The `PARAMETERS FROM CALLBACK BODY` should be filled with the urlencoded body yo
 
 Furthermore, when handling the incoming credentials on the client, make sure to only overwrite the current (guest) session of the user once your own server have validated the incoming `code` parameter, such that your app is not susceptible to malicious incoming links (e.g. logging out the current user).
 
-### iOS
\ No newline at end of file
+### iOS
+
```
HenriBeck, line 166:
```suggestion
At this point, you should have added the Sign in with Apple capability to either your own app's capabilities or the test application you created to run the example.

In case you don't have `Automatically manage Signing` turned on in XCode, you will need to recreate your Provisioning Profiles for your App, so they include the new `Sign in with Apple` capability. Then you can download the new certificates and select them in XCode.
In case XCode manages your signing, this step will be done automatically for you.

Additionally, this assumes that you have at least one iOS device registered in your developer account for local testing.
```

What do you think about this? I think it's more friendly to users who don't manage their certificates through XCode.

tp:
The hardest part here was figuring out that the obscure build errors were due to `FLUTTER_ROOT` not being set due to a wrong build config (that I changed in response to the Cocoapods warning)… Well, in the end worth it to do a from-scratch setup in a clean test account.

packages/sign_in_with_apple/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -374,7 +378,7 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/Flutter",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.example.signInWithAppleExample;
+				PRODUCT_BUNDLE_IDENTIFIER = de.lunaone.flutter.signinwithappleexample;
```
HenriBeck, line 381: Should we maybe create an ay specific bundle identifier that we can just use with our AY addresses?

packages/sign_in_with_apple/example/android/app/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/example/MainActivity.kt
```diff
@@ -0,0 +1,6 @@
+package com.aboutyou.dart_packages.sign_in_with_apple.example
```
HenriBeck, line 1: I like this bundle id more. Let's remove the other files and change the example app to use this one?

HenriBeck: Can also be a quick follow up to get this merged

packages/sign_in_with_apple/example/ios/Runner.xcodeproj/project.pbxproj
tp: Especially with the automatic signing, I didn't want to auto-update anything in our production account right from Xcode, so I used another developer account for testing.

If you think we can create a test app and share the certificates manually, that would be great. We will definitely have to start this a bunch of times in the future for debugging.

tp: Follow up @ https://github.com/aboutyou/dart_packages/issues/66

Issue #62, opened 2020-04-27
Integration with Firebase Auth


miguelpruivo:
Hi,

Thank you for the package. I need to integrate this with Firebase Auth (using oAuth2) and unfortunately, the `AuthorizationCredentialApple` doesn't have the `idToken` that I need in order to get a valid `AuthCredential`. 

Do you know a way to get it or isn't currently unsupported?

Thanks!

HenriBeck:
In practice, this should be possible.

I just saw that we are not exposing the `identityToken` anymore. I think we didn't see an immediate use case before, but we can definitely add the token back.

I guess we would also need to support the `nonce` for all platforms to enable Firebase Auth to work.

But other than that this should be possible.

miguelpruivo:
@HenriBeck thanks for the fast response. Any clue of when this might be added?

Thanks.

HenriBeck:
I think adding those two features shouldn't be hard. I will provide two PRs now, and they should be released until EOW

HenriBeck:
The PRs have been created. I only need to check android if we are getting the identity token there

HenriBeck:
@miguelpruivo I just validated that the identity token works on Android, so I think both PRs are finished. We will probably not be able to test the integration with Firebase ourselves at the moment.

miguelpruivo:
@HenriBeck no problem, I’ll test it myself on iOS side and let you know. Thanks for the quick implementation!

HenriBeck:
@miguelpruivo We have released version 2.1.0 with `nonce` and `identityToken` support.

This should be all the work that is needed on our site; therefore I will close this ticket. If you encounter any other integration issues, please let us know.

miguelpruivo:
Thank you @HenriBeck!

tp:
@miguelpruivo Did this solve your issue? In #73 there is another question of how to integrate Firebase auth, so it would be super helpful if you could share your code if it's working for you now, so we could add that as an example here.

Thanks 🙏

PR #63, opened 2020-04-27
Expose identity token again

HenriBeck:
#62


packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -197,7 +197,7 @@ class SignInWithApple {
             .join(' '),
         // Request `code`, which is also what `ASAuthorizationAppleIDCredential.authorizationCode` contains.
         // So the same handling can be used for Apple and 3rd party platforms
-        'response_type': 'code',
+        'response_type': 'code+id_token',
```
HenriBeck, line 200: @tp Do you think it makes sense to configure this from the outside?

HenriBeck: FYI: I added the `id_token` response param according to https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms#3332113

The `code+id_token` has been taken from https://forums.developer.apple.com/thread/118209. I still need to validate this

HenriBeck: Update: I tried the android code and it seems to be `code id_token` in the query param so I adjusted the code for this.

tp:
@HenriBeck Do you have any insight on how this works? It says that the identity token is basically the JWT token, which is (to the best of my knowledge) also what you get back when you ask Apple with the `code`. So why would one prefer to use the identity token instead?

(Wondering whether this would also simplify anything for us.)

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
tp: 👍

PR #64, opened 2020-04-27
Add nonce support

HenriBeck:


packages/sign_in_with_apple/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 2.1.0
 
 - Expose `identityToken` to enable Firebase integration (https://github.com/aboutyou/dart_packages/issues/62)
+- Add support for passing a `nonce` to the authentication request
```
tp, line 4: @HenriBeck Extended the current CHANGELOG (and did not publish yet), as this should go together with the previous feature if I understood correctly, right?

packages/sign_in_with_apple/lib/src/authorization_request.dart
```diff
@@ -54,13 +55,19 @@ class AppleIDAuthorizationRequest implements AuthorizationRequest {
   /// For more information see: https://forums.developer.apple.com/thread/121496
   final List<AppleIDAuthorizationScopes> scopes;
 
+  /// A String value used to associate a client session with an ID token. This value is also used to mitigate replay attacks.
```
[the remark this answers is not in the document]
tp: Okay, so either one would use a timestamp, or something the server can ensure is only used once (e.g. it stores all previous nonces, or uses a counter like "login attempt", and then denying everything that comes in with a lower number).

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -82,6 +82,11 @@ class SignInWithApple {
     ///
     /// This parameter is required on Android.
     WebAuthenticationOptions webAuthenticationOptions,
+
+    /// A String value used to associate a client session with an ID token. This value is also used to mitigate replay attacks.
+    ///
+    /// Can be `null`, in that case no nonce will be passed to the request.
```
tp, line 88:
```suggestion
    ///  Optional string which, if set, will be be embedded in the resulting `identityToken` field on the [AuthorizationCredentialAppleID].
    ///     
    /// This can be used to mitigate replay attacks by using a unique argument per sign-in attempt.
    ///
    /// Can be `null`, in which case no nonce will be passed to the request.
```

tp: What do you think of this?

I would then also do a specialized version for the output value.

packages/sign_in_with_apple/lib/src/authorization_request.dart
tp, line 58:
```suggestion
  /// The nonce value which was provided when initiating the sign-in.
```

packages/sign_in_with_apple/lib/src/authorization_request.dart
```diff
@@ -54,13 +55,19 @@ class AppleIDAuthorizationRequest implements AuthorizationRequest {
...
+  ///
+  /// Can be `null`
```
tp, line 60:
```suggestion
  /// Can be `null` if no value was given on the request.
```

packages/sign_in_with_apple/lib/src/authorization_request.dart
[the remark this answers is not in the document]
HenriBeck: I was mostly following the requirements which are shown here: https://firebase.google.com/docs/auth/ios/apple

The `nonce` is on the `OAuthRequest` which the `AppleIDRequest` inherits from.

packages/sign_in_with_apple/CHANGELOG.md
HenriBeck: Yeah, both features are needed to allow integration with Firebase Auth.

packages/sign_in_with_apple/lib/src/authorization_request.dart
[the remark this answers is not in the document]
tp: Ah, interesting. I would've thought using it through Firebase would be fully integrated. So that's why one still needs a plugin like this.

PR #68, opened 2020-04-28
Run `flutter format`

tp:


tp:
"Health suggestion" from https://pub.dev/packages/sign_in_with_apple#-analysis-tab-

PR #69, opened 2020-04-28
v2.1.0+1

HenriBeck:


packages/sign_in_with_apple/example/pubspec.lock
```diff
@@ -1,48 +1,62 @@
 # Generated by pub
 # See https://dart.dev/tools/pub/glossary#lockfile
 packages:
+  archive:
```
HenriBeck, line 4: @tp what flutter version do you have?

tp: Same, when I run `flutter packages get`:

```
  archive:
    dependency: transitive
    description:
      name: archive
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.0.11"
  args:
    dependency: transitive
    description:
      name: args
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.5.2"
```

packages/sign_in_with_apple/pubspec.lock
```diff
@@ -120,7 +162,7 @@ packages:
       name: test_api
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.2.15"
+    version: "0.2.11"
```
tp, line 165: Also this gets downgraded on the AY user (Flutter 1.12), compared to my fresh dev account which is using `flutter dev` channel (as I need macOS support there).

PR #70, opened 2020-04-30
Purge test team and bundle ID from example project

tp:


packages/sign_in_with_apple/example/pubspec.lock
```diff
@@ -1,48 +1,62 @@
 # Generated by pub
 # See https://dart.dev/tools/pub/glossary#lockfile
 packages:
+  archive:
```
tp, line 4: These changed again as I now used the stable 1.12 release instead of the dev channel…

Issue #81, opened 2020-05-22
[sign_in_with_apple] On Android, the app is restarting after the callback intent from oauth redirect is received


eduribas:
On Android, after calling `await SignInWithApple.getAppleIDCredential()` and performing the authentication on the chrome tab, the call returns with the credential but the app is restarted. If I put a breakpoint on `runApp(MyApp());` I can see that it's called again. Is that the expected behavior?

I was able to reproduce that using the example app.

HenriBeck:
Hey @eduribas, I will have a look at this tomorrow.

Could you please post the output of `flutter doctor -v` so I can know on which versions you are?

eduribas:
Hi @HenriBeck , I tracked down the problem and sent a PR that seems to solve the problem on my case. This is my flutter doctor -v:
```
[✓] Flutter (Channel stable, v1.17.1, on Mac OS X 10.15.4 19E287, locale en-BR)
    • Flutter version 1.17.1 at /Users/eduribas/Dev/Library/flutter
    • Framework revision f7a6a7906b (10 days ago), 2020-05-12 18:39:00 -0700
    • Engine revision 6bc433c6b6
    • Dart version 2.8.2

 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.1)
    • Android SDK at /Users/eduribas/Library/Android/sdk
    • Platform android-29, build-tools 29.0.1
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 11.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 11.4.1, Build version 11E503a
    • CocoaPods version 1.8.4

[✓] Android Studio (version 3.6)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 45.1.1
    • Dart plugin version 192.7761
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)

[✓] VS Code (version 1.45.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.10.2

[✓] Connected device (1 available)
    • SM G975F • RQ8M20K40ZF • android-arm64 • Android 10 (API 29)

[...]

PR #82, opened 2020-05-22
Fix Android main activity is recreated after processing the intent callback

eduribas:
I'm not sure if the call to close the chrome custom tab using FLAG_ACTIVITY_NEW_TASK + FLAG_ACTIVITY_RESET_TASK_IF_NEEDED flags was intended, but it was causing the flutter app to restart. Changing it to use FLAG_ACTIVITY_CLEAR_TOP solves the problem since it closes the chrome custom tab without recreating the main activity. This is related to the issue #81


tp:
To be honest I was not entirely confident with the flags, and was just looking for a way that the app starting the Chrome custom tab would be brought back to the foreground, closing / popping the Chrome Custom Tab.

I'll check this out locally quickly, code wise the change looks good.

Thanks for your contribution :)

tp:
With the current approach (from v2.2) it indeed calls `runApp` again (seemingly after the credentials have been passed to the app; see `credential`, `session`, `init` order in the log):

image

For me the change in this PR doesn't work though, it keeps hanging in the Chrome Custom Tab during login (not sure if it's an issue with the server redirect currently, but `master` worked just a few minutes ago).

tp:
Hmm, seems like there was a glitch in my setup / getting no response from the server.

Now I was able to have multiple successful runs with the changed code on the example, all without a new activity being created :)

image


Thanks for you contribution @eduribas, we'll be releasing this as 2.3.0.

PR #87, opened 2020-05-29
Bump Flutter version to be used

### LINKED ISSUE #86: [sign_in_with_apple] Make fallback font configurable
We should make the fallback font configurable for the SiwA button, so app's can provide their own font name for non-Apple platforms (where the SF font is not available).

tp:
Closes #86


tp:
@HenriBeck I'll fix this up today, so we can use that for #88

PR #88, opened 2020-06-01
[sign_in_with_apple] Handle cancellation of chrome custom tab

HenriBeck:


packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -110,7 +118,23 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
...
-    activity = null
+    binding?.removeActivityResultListener(this)
+    binding = null
+  }
+
+  override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?): Boolean {
+    if (requestCode == CUSTOM_TABS_REQUEST_CODE) {
+      val _lastAuthorizationRequestResult = lastAuthorizationRequestResult
+
+      if (_lastAuthorizationRequestResult != null) {
```
HenriBeck, line 129: While this works from my testing on the simulator, I'm a bit worried about introducing a race condition here.

But maybe this activity result callback is only triggered after the deeplink is handled, e.g. triggering the main activity to hide the Custom Tab

PR #93, opened 2020-06-07
[secure-storage] Initial plugin

HenriBeck:
This implements a common interface for secure storage for Android, iOS and macOS.

On iOS, we are using the standard keychain service to store this data.

For Android, we have not yet decided how to store the data


packages/secure_storage/ios/Classes/SwiftSecureStoragePlugin.swift
```diff
@@ -0,0 +1,71 @@
...
+        return
+    }
+    
+    switch call.method {
+    case "read":
+        var query = parseFlutterMethodQuery(args: args)
+        query[kSecReturnData] = true
+        
+        var item: CFTypeRef?;
+        let status = SecItemCopyMatching(query as CFDictionary, &item)
```
HenriBeck, line 31: @tp I had a look at the secure storage for flutter, and the keychain implementation might not be as trivial as we have thought, especially if we want to support all options.

The Keychain also behaves more like a database and not like a key/value storage.

All options are technically documented here: https://developer.apple.com/documentation/security/keychain_services/keychain_items/item_attribute_keys_and_values
Note that there are also different types of keychain items, e.g. a generic password or an internet password. And technically you don't even need a type, but you could also just save raw data in the keychain.

The `flutter-secure-storage` seems to abstract all of this by treating every value as a generic password and making the key the `kSecAttrAccount` value.

tp: @HenriBeck My thought for a package like this was to first focus on the "website password" use case. So that we can store and retrieve passwords for associated domains.

That should be the most common scenario in keychain and is also finally available [Smart Lock for Passwords on Android](https://developers.google.com/identity/smartlock-passwords/android).

What do you think about that scope? Especially as it gives nice feature parity between iOS and Android, as well as syncing through the cloud (if the user has it enabled).

PR #98, opened 2020-06-11
SiwA 2.5.0

HenriBeck:


packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -156,12 +156,12 @@ AuthorizationCredentialAppleID parseAuthorizationCredentialAppleIDFromDeeplink(
   Uri deeplink,
 ) {
   if (deeplink.queryParameters.containsKey('error')) {
-    // In case the user canceled the sign-in, the URL will have an `error` parameter
-    // 
-    // The only error code that might be returned is `user_cancelled_authorize`, 
-    // which indicates that the user clicked the `Cancel` button during the web flow
-    //
-    // https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms
+    /// In case an error occured during the web flow, the URL will have an `error` parameter.
```
tp, line 159: I guess this was fine (with 2 `/`) as it's not a doc comment. But don't mind…

HenriBeck: Yeah, the `///` comments just look better in the IDE imo

PR #105, opened 2020-06-25
Added Semantics widget to SignInWithAppleButton

AndroidNils:
The `Semantics` widget is used by accessibility tools, search engines, and other semantic analysis software to determine the meaning of the application (for more information: https://api.flutter.dev/flutter/widgets/Semantics-class.html). So I added this to the `SignInWithAppleButton` to have this out-of-the-box support 👍

### DIFF
--- a/packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
+++ b/packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
@@ -166,25 +166,31 @@ class SignInWithAppleButton extends StatelessWidget {
         break;
     }
 
-    return Container(
-      height: height,
-      child: SizedBox.expand(
-        child: CupertinoButton(
-          borderRadius: borderRadius,
-          padding: EdgeInsets.zero,
-          color: _backgroundColor,
-          child: Container(
-            decoration: _decoration,
-            padding: EdgeInsets.symmetric(
-              horizontal: 16.0,
+    return Semantics(
+      button: true,
+      onTap: onPressed,
+      label: text,
+      enabled: onPressed != null,
+      child: Container(
+        height: height,
+        child: SizedBox.expand(
+          child: CupertinoButton(
+            borderRadius: borderRadius,
+            padding: EdgeInsets.zero,
+            color: _backgroundColor,
+            child: Container(
+              decoration: _decoration,
+              padding: EdgeInsets.symmetric(
+                horizontal: 16.0,
+              ),
+              height: height,
+              child: Row(
+                children: children,
+                mainAxisAlignment: MainAxisAlignment.center,
+              ),
             ),
-            height: height,
-            child: Row(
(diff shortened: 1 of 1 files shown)


packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
```diff
@@ -166,25 +166,31 @@ class SignInWithAppleButton extends StatelessWidget {
...
-          child: Container(
-            decoration: _decoration,
-            padding: EdgeInsets.symmetric(
-              horizontal: 16.0,
-            ),
-            height: height,
-            child: Row(
-              children: children,
-              mainAxisAlignment: MainAxisAlignment.center,
+    return Semantics(
```
HenriBeck, line 169: @AndroidNils The `CupertinoButton` already adds a `Semantics` widget, but it seems to not have `label` for example and only `button: true`.

For what is this `Semantics` widget exactly needed?

AndroidNils: Like I mentioned in the PR description the `Semantics` widget is for accessibility tools, search engines, and other semantic analysis software. To the best of my knowledge, "label" is used for Screen Reader to indicate the meaning of a button.

AndroidNils: But I see that the Text Widget has also a `Semantics` Widget. I don't know, if screen reader can mix the information from the button and the text of the button. I will research on this 👍

AndroidNils: So regarding to the `Semantics` widget [docs](https://api.flutter.dev/flutter/widgets/Semantics-class.html) ("A widget that annotates the widget tree with a description of the meaning of the widgets") Flutter should merge the `Semantics` of the `CupertinoButton` and the `Text` widgets, because they are in one widget tree. Nevertheless I was still surprised why the CupertinoButton Semantics widget did not pass the `onTap` method, because the `Semantics.onTap` is used VoiceOver users on iOS and TalkBack users on Android can trigger this action by double-tapping the screen while an element is focused. The reason for this is that the `CupertinoButton` uses the `GestureDetector`, which pass the parameter to the `Semantics` widget.

Therefore `label` and `onTap` should already be implemented by Flutter. Only `enabled` could be added. But actually it would make more sense, if `enabled` would be added to the `Semantics` widget of `CupertinoButton` (under `button: true`).

I just tried our App Sharezone with TalkBalk (Screen Reader for Android) again. Apple Sign In Button is wonderfully recognized 👍  So I am going to close this PR.

tp: Thanks for the thorough explanation and looking into great accessibility support for your app and this widget :)

PR #113, opened 2020-07-12
[sign in with apple] Add generateNonce util function

HenriBeck:


packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -145,32 +168,61 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
  *
  * DO NOT rename this or it's package name as it's configured in the consumer's `AndroidManifest.xml`
  */
-public class SignInWithAppleCallback: Activity {
-  constructor() : super()
+public class SignInWithAppleCallback : Activity() {
+  private fun validateNonce( uri: Uri, expectedNonce: String?): Boolean {
```
tp, line 172:
```suggestion
  private fun validateNonce(uri: Uri, expectedNonce: String?): Boolean {
```

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -145,32 +168,61 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
...
+    if (idToken == null) {
+      Log.e(TAG, "Missing id_token query parameter in signinwithapple callback")
+      return false
+    }
+
+    try {
+      val jwt = JWT(idToken)
+
+      if (jwt.getClaim("nonce_supported").asBoolean() == true) {
+        if (jwt.getClaim("nonce").asString() != expectedNonce) {
```
tp, line 183: I guess this warrants a comment, that we don't want to early return in case of the `expectedNonce` being `null` as we still check that the incoming data also doesn't have one.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -145,32 +168,61 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
...
     }
 
-    val triggerMainActivityToHideChromeCustomTab = SignInWithApplePlugin.triggerMainActivityToHideChromeCustomTab
-    if (triggerMainActivityToHideChromeCustomTab != null) {
-      triggerMainActivityToHideChromeCustomTab()
-      SignInWithApplePlugin.triggerMainActivityToHideChromeCustomTab = null
-    } else {
-      Log.e(TAG, "Received Sign in with Apple callback, but 'triggerMainActivityToHideChromeCustomTab' function was `null`")
+    if (!validateNonce(uri, onGoingLoginAttempt.nonce)) {
+      Log.e(TAG, "Received Sign in with Apple callback, but the nonce parameter validation failed")
```
tp, line 214:
```suggestion
      // Just log the occurrence, but don't affect any ongoing logins (so the app can't be spammed)
      Log.e(TAG, "Received Sign in with Apple callback, but the nonce parameter validation failed")
```

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
HenriBeck: I wouldn't add any default nonce as this might lead to unexpected login failures in case someone tries to validate it (Firebase does, for example).

I would instead provide a small function for generating this nonce so the users don't have to write it themselves.

tp:
```suggestion
        // Check that the given nonce matched the expected one, even if none is expected
        if (jwt.getClaim("nonce").asString() != expectedNonce) {
```

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -145,32 +168,62 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
...
+
+    try {
+      val jwt = JWT(idToken)
+
+      if (jwt.getClaim("nonce_supported").asBoolean() == true) {
+        if (jwt.getClaim("nonce").asString() != expectedNonce) {
+          Log.e(TAG, "Expected that the JWT nonce matches the initially provided one, but they differed")
+          return false
+        }
+      }
```
tp, line 187:
```suggestion
      } else if (expectedNonce != null) {
          Log.e(TAG, "Expected a nonce on the token, but none was included")
          return false
      }
```

I think this case was not handled yet. If the token didn't have a nonce, it wouldn't run into the detailed check in the first place.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
tp: Hmm, with the additional check below (https://github.com/aboutyou/dart_packages/pull/113/files#r464235837), I don't think we would even go into this branch if the expected nonce was `null`. So that comment doesn't make sense.

HenriBeck:
@tp For now, I removed the code around parsing the JWT mainly because the server should do the nonce validation once it has verified that the JWT token comes from Apple.

For now, this PR only adds a function to generate a cryptographically secure nonce.

tp (approved):

Issue #85, opened 2020-05-26
sign_in_with_apple: MISSING_ACTIVITY on android


Dreamersoul:
Hello, the plugin works perfectly on iOS for me however i added the web parameters
```dart
await SignInWithApple.getAppleIDCredential(
    scopes: [
        AppleIDAuthorizationScopes.email,
        AppleIDAuthorizationScopes.fullName,
     ],
    webAuthenticationOptions: WebAuthenticationOptions(
    clientId: "com.example.app",
    redirectUri: Uri.parse("https://${Constants.BASE_HOST}/user/auth/apple/callback")
);
```
however when clicking on the button i get the following error:
```logcat
2020-05-26 16:45:26.864 7062-7114/com.exampe.app E/flutter: [ERROR:flutter/lib/ui/ui_dart_state.cc(157)] Unhandled Exception: PlatformException(MISSING_ACTIVITY, Plugin is not attached to an activity, {url: https://appleid.apple.com/auth/authorize?client_id=com.example.app&redirect_uri=https%3A%2F%2Fexample.herokuapp.com%2Fuser%2Fauth%2Fapple%2Fcallback&scope=email+name&response_type=code+id_token&response_mode=form_post})
    #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:569:7)
    #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:156:18)
    <asynchronous suspension>
    #2      MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:329:12)
    #3      SignInWithApple._signInWithAppleAndroid (package:sign_in_with_apple/src/sign_in_with_apple.dart:231:34)
    #4      SignInWithApple.getAppleIDCredential (package:sign_in_with_apple/src/sign_in_with_apple.dart:107:14)
    #5      _RegisterWidgetState.buildRegisterButtons.<anonymous closure> (package:example/widgets/profile/register/register_widget.dart:219:56)
    #6      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:779:19)
    #7      _InkResponseState.build.<anonymous closure> (package:flutter/src/material/ink_well.dart:862:36)
    #8      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:182:24)
    #9      TapGestureRecognizer.handleTapUp (package:flutter/src/gestures/tap.dart:504:11)
[...]

PR #119, opened 2020-07-27
Support both v1 and v2 Android Embeddings

jaggernod:
The plugin now can be registered multiple times with the v1 Android Embeddings.

Most of the dependencies added are for E2E testing of both embeddings.
Have also added a README with an explanation.


tp:
@jaggernod Thanks for your contribution.

To be honest I am not super familiar with the embeddings, especially the v1 version, so we'll have to be careful if we take this on.

Out of curiosity, what is preventing you from upgrading to V2 - Is there any chance that dependency can be moved forward instead? (related https://github.com/aboutyou/dart_packages/issues/85#issuecomment-662328796)

jaggernod:
@tp, unfortunately, moving of a medium-size project to v2 is quite hard especially when one has plenty of their own plugins. 
In the worst scenario, we can just use your fork with the path for the time being.

HenriBeck:
@jaggernod The V2 embedding is backward compatible. So you don't need to migrate all plugins at once. We also still have quite a few third-party plugins using the V1 embedding.

jaggernod:
You are totally right @HenriBeck. But for us the biggest issue is background processing. 
Firebase Messaging plugin still requires registration in Application (as it needs to exist when no Activity exists). It is then when registering your plugin fails.

We have our own plugin for communication from background job which faces the same problem.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -53,62 +56,91 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
...
+        override fun onActivityResumed(activity: Activity) = Unit
+        override fun onActivityDestroyed(activity: Activity?) = plugin.unregisterActivity()
+        override fun onActivitySaveInstanceState(activity: Activity?, outState: Bundle?) = Unit
+        override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) = plugin.registerActivity(activity)
+        override fun onActivityStarted(activity: Activity) = Unit
+        override fun onActivityStopped(activity: Activity) = Unit
+      })
     }
+
+    private fun application(registrar: Registrar) = registrar.context().applicationContext as Application
```
HenriBeck, line 79: If I'm not mistaken, these `application` and `registerLifecycle`  are not used anywhere else. What do you think about inlining them to keep the deprecated support in a single function?

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -53,62 +56,91 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
...
+  }
+
+  fun registerActivity(currentActivity: Activity) {
+    activity = currentActivity
```
HenriBeck, line 83: Is there anything against directly setting the activity on the plugin instead of this function?

packages/sign_in_with_apple/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -37,6 +37,15 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
             </intent-filter>
         </activity>
+
+        <activity
```
HenriBeck, line 41:
```suggestion
        <!-- An addtional activity for testing Android Embedding V1 -->
        <activity
```

packages/sign_in_with_apple/example/android/app/build.gradle
```diff
@@ -62,6 +62,8 @@ flutter {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     testImplementation 'junit:junit:4.12'
-    androidTestImplementation 'androidx.test:runner:1.1.1'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.1.1'
+    testImplementation 'junit:junit:4.12'
```
HenriBeck, line 65:
```suggestion
```

HenriBeck:
@jaggernod It would be great if you could merge our latest master changes, then the CI would also be run for this PR.

jaggernod:
@HenriBeck I have a question. 
Were you able to test the plugin on Android device? After (and before) my changes I was always having problems to successfully sign in. 
So actually I dripped support for Android.

HenriBeck:
@jaggernod yeah, for us the sign-in on Android is working.

jaggernod:
Thanks, will apply the changes when I'm back from holidays. First week of September.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
jaggernod: Have done it this way for a better readability. But will revert.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
jaggernod: I'm a fun of small functions. But understand that you want to at one time just wype everything at once.
Will change.

packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
jaggernod: I did not want to make the `application` public. Is it OK with you or should I expose it?

DRSchlaubi:
I don't see a reason for supporting v1. V2 has the ShimRegistry which allows you to support v1 plugins in v2 embedded apps

jaggernod:
Ok, no problem

Issue #125, opened 2020-08-06
[sign_in_with_apple] Task :sign_in_with_apple:compileDebugKotlin FAILED


HugoSart:
Unable to build app after importing the package on pubspec with the following error:
```
> Task :sign_in_with_apple:compileDebugKotlin
e: C:\Users\hugo_\AppData\Roaming\Pub\Cache\hosted\pub.dartlang.org\sign_in_with_apple-2.5.2\android\src\main\kotlin\com\aboutyou\dart_packages\sign_in_with_apple\SignInWithApplePlugin.kt: (86, 29): Only safe (?.) or non-null asserted (!!.) calls are allowed on a n
ullable receiver of type Intent?
e: C:\Users\hugo_\AppData\Roaming\Pub\Cache\hosted\pub.dartlang.org\sign_in_with_apple-2.5.2\android\src\main\kotlin\com\aboutyou\dart_packages\sign_in_with_apple\SignInWithApplePlugin.kt: (88, 29): Only safe (?.) or non-null asserted (!!.) calls are allowed on a n
ullable receiver of type Intent?
```

hugopretorius914:
I get the same error, any ideas?

HenriBeck:
This has been fixed this week @hugopretorius914, just upgrade to the latest version which is 2.5.4.

PR #126, opened 2020-08-06
Added missing null checkers

HugoSart:
Related to #125

### DIFF
--- a/packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
+++ b/packages/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
@@ -83,9 +83,9 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
         lastAuthorizationRequestResult = result
         triggerMainActivityToHideChromeCustomTab = {
           val notificationIntent = _activity.packageManager.getLaunchIntentForPackage(_activity.packageName);
-          notificationIntent.setPackage(null)
+          notificationIntent?.setPackage(null)
           // Bring the Flutter activity back to the top, by popping the Chrome Custom Tab
-          notificationIntent.flags = Intent.FLAG_ACTIVITY_CLEAR_TOP;
+          notificationIntent?.flags = Intent.FLAG_ACTIVITY_CLEAR_TOP;
           _activity.startActivity(notificationIntent)
         }
 


HenriBeck:
@HugoSart thanks for the quick fix. Could you please quickly provide the flutter version you are on and the kotlin version to see when this breaks.

HugoSart:
@HenriBeck sure.
Flutter: 1.20.1
Kotlin: 1.3

HenriBeck:
@HugoSart could you maybe provide a repository where this issue appears? 

I created a new flutter project with version 1.20 and there the compilation works. 
Also upgrading our example app to flutter 1.20.1 and it runs on android without problems.

HugoSart:
@HenriBeck unfortunately I was not able to reproduce this problem by starting from a fresh project.

I created a new project, copied the pubspec.yml, and both build.gradle from the broken project to the working one, but the problem did not happen. I'm exploring it more trying to find the reason.

OBS: I should've mentioned it before, but this problem only happens when building with --release option.

HenriBeck:
@HugoSart does this maybe happen with some special android sdk version? Either the minSdkVersion or targetSdkVersion?

hespinola:
I'm having this exact same problem.

Flutter version 1.20.4
Android minSdkVersion 21
Android targetSdkVersion 29
Kotlin version 1.4.10

HugoSart:
This was some time ago, and as I remember, the solution was to add these lines to the root build.gradle:
```
buildscript {
    ext.kotlin_version = '1.3.72'
    // other definitions ...

    dependencies {
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
       // other definitions ...
    }

}
```
So, basically, I set the kotlin version to 1.3.72. As I said, this is what I remember, maybe I did something else too, but you could give it a try @hespinola .

wahyubucil:
I'm having the same problem. Changing the Kotlin version to 1.3.72 doesn't help

Flutter version 1.22.1
Android minSdkVersion 21
Android targetSdkVersion 29
Kotlin version 1.3.72

hespinola:
> I'm having the same problem. Changing the Kotlin version to 1.3.72 doesn't help
> 
> Flutter version 1.22.1
> Android minSdkVersion 21
> Android targetSdkVersion 29
> Kotlin version 1.3.72

These are my currently working settings:

targetSdkVersion 30
Kotlin version 1.3.50

Then inside build.gradle:

classpath 'com.android.tools.build:gradle:3.6.4'

And in gradle-wrapper.properties:

distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip

wahyubucil:
@hespinola The problem still exists. I already use those settings

I check the `SignInWithApplePlugin.kt` on Android Studio. And it shows error too. So I think we need to merge this PR

yswtrue:
Same issue

hespinola:
I can confirm the problem still exists, I mentioned it worked in my previous comment but I didn't realize I was using the forked version with the fix.

@tp @HenriBeck  If this PR solves the issue and doesn't seem to break the code I don't see why it wouldn't be merged.

HenriBeck:
@hespinola we will have a look at this now that this affects more people.

tp:
@hespinola You're right in that this seems harmless, but we also want to find out the reason for this to maintain it properly in the future.

After upgrading to 

* `ext.kotlin_version = '1.4.10'`
* `classpath 'com.android.tools.build:gradle:3.6.4'`
* `distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip` 

I did not run into any build issues with `sign_in_with_apple` 2.5.2 using `release` builds in CI.

So I think there is some other difference between the projects that run into this issue and those who don't.

tp:
I found the way to run into this issue:

Setting `compileSdkVersion 29` in `packages/sign_in_with_apple/android/build.gradle`.

Which would make sense if the nullability annotations where updated in that one.

My assumption is that somehow this gets overwritten for some projects, but not others, hence only few run into this. 

Anyway, I tried @HugoSart's fix locally in the then failing example project, and this indeed fixed the issue.

Will release an update shortly.

tp (approved):

tp:
Thanks @HugoSart & @hespinola for your sustained effort in getting this fixed :)

PR #143, opened 2020-09-18
Add `state_queue_test` package

tp:
cc @wwwdata

The check in `bloc`'s `blocTest` helper to alert the user if their state type is missing `Equatable` conformance is also quite nice: https://github.com/felangel/bloc/blob/master/packages/bloc_test/lib/src/bloc_test.dart#L175

Definitely something the AY app's states are mostly missing at the moment, so we'd benefit there from using generated state classes as well.


packages/state_queue_test/lib/src/state_queue_test.dart
```diff
@@ -0,0 +1,62 @@
...
+import 'package:meta/meta.dart';
+import 'package:state_queue/state_queue.dart';
+import 'package:test/test.dart' as test;
+
+@isTest
+void blocTest<B extends StateQueue<State>, State>(
+  String description, {
+  @required B Function() build,
+  FutureOr<void> Function(B bloc) act,
+  Iterable<State> expect,
```
HenriBeck, line 13:
```suggestion
  Matcher expect,
```

I would keep this more "open", it would still allow you to use `equals([...])` but also allow something like `isA<>` matcher.

packages/state_queue_test/test/state_queue_test/implementation_test.dart
```diff
@@ -0,0 +1,92 @@
...
+        );
+      },
+      // ignore: avoid_annotating_with_dynamic
+      throwsA((dynamic e) =>
+          e is TestFailure &&
+          e.message.contains('Which: longer than expected')),
+    );
+  });
+
+  test('Errors with too wrong states', () {
```
HenriBeck, line 75:
```suggestion
  test('Errors with wrong states', () {
```

packages/state_queue_test/test/state_queue_test/implementation_test.dart
```diff
@@ -0,0 +1,92 @@
...
+        );
+      },
+      // ignore: avoid_annotating_with_dynamic
+      throwsA((dynamic e) =>
+          e is TestFailure &&
+          e.message.contains('Which: shorter than expected')),
+    );
+  });
+
+  test('Errors with too many states', () {
```
HenriBeck, line 57:
```suggestion
  test('Errors with too few states', () {
```

packages/state_queue_test/test/state_queue_test/implementation_test.dart
```diff
@@ -0,0 +1,92 @@
...
+          act: (b) {
+            b.incr();
+          },
+          expect: [0, 7],
+        );
+      },
+      // ignore: avoid_annotating_with_dynamic
+      throwsA((dynamic e) =>
+          e is TestFailure &&
+          e.message.contains('Which: was <1> instead of <7> at location [1]')),
```
HenriBeck, line 89:
```suggestion
          e.message.contains('Which: was <1> instead of <7> at location [1]'),),
```

packages/state_queue_test/test/state_queue_test/implementation_test.dart
```diff
@@ -0,0 +1,92 @@
...
+          act: (b) {
+            b..incr()..incr()..incr();
+          },
+          expect: [0, 1],
+        );
+      },
+      // ignore: avoid_annotating_with_dynamic
+      throwsA((dynamic e) =>
+          e is TestFailure &&
+          e.message.contains('Which: longer than expected')),
```
HenriBeck, line 71:
```suggestion
          e.message.contains('Which: longer than expected'),),
```

packages/state_queue_test/test/state_queue_test/implementation_test.dart
```diff
@@ -0,0 +1,92 @@
...
+        await runBlocTest<_TestBloc, int>(
+          build: () => _TestBloc(),
+          // act: (_) => throw Exception('Act failure'),
+          expect: [0, 1],
+        );
+      },
+      // ignore: avoid_annotating_with_dynamic
+      throwsA((dynamic e) =>
+          e is TestFailure &&
+          e.message.contains('Which: shorter than expected')),
```
HenriBeck, line 53:
```suggestion
          e.message.contains('Which: shorter than expected'),),
```

wwwdata (approved):
nice first implementation. My only complaint is also to accept matchers instead of just states.

packages/state_queue_test/lib/src/state_queue_test.dart
tp: I think the difference between `Iterable` (which would be states or `Matcher`s) vs a single `Matcher` is important though.

My personal guess is that you mostly want state item level checks, and not so much globals (like `isEmpty`, `containsMoreThan(5)`, or similar).

So `Iterable<Matcher|State>` or a single `Matcher`?

packages/state_queue_test/test/state_queue_test/implementation_test.dart
```diff
@@ -0,0 +1,92 @@
...
+      throwsA((dynamic e) => e.toString() == 'Exception: Act failure'),
+    );
+  });
+
+  test('Errors with too many states', () {
+    expect(
+      () async {
+        await runBlocTest<_TestBloc, int>(
+          build: () => _TestBloc(),
+          // act: (_) => throw Exception('Act failure'),
```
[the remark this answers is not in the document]
tp:
```suggestion
```

packages/state_queue_test/lib/src/state_queue_test.dart
wwwdata: Just `Iterable` because that is what works in the end with the `expect` function. And we don't want to take away functionality, right?

packages/state_queue_test/lib/src/state_queue_test.dart
```diff
@@ -0,0 +1,62 @@
...
+  void collectState() {
+    states.add(bloc.value);
+  }
+
+  if (collectStates) {
+    bloc.addListener(collectState);
+  }
+
+  if (collectStates) {
+    states.add(bloc.value);
```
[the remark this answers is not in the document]
tp: Interesting. I would also think it makes sense to test the initial state if there are any parameters to it. Else one would have to setup the bloc twice with the same data. Not sure if that is not too granular.

I think this is what `skip` is for, which for some kinds of blocs (which boring / static loading states), might just be the thing to do (though personally I think I would rather assert on the loading instead of skipping it, both of which are 1 line).

For reference here's `bloc_test` (which also uses `setUp` to "build" the bloc instead of doing it inline):

image


image

packages/state_queue_test/lib/src/state_queue_test.dart
tp: > Just `Iterable` because that is what works in the end with the `expect` function. And we don't want to take away functionality, right?

But then you can't use top-level matchers like `isEmpty`, and also need not use `equals([…])` as shown above.

Which is fine I guess, as I think asserting on the individual states is the focus for this kind of test.

packages/state_queue_test/lib/src/state_queue_test.dart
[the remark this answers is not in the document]
wwwdata: yes, this is what I meant. From this example, you can clearly see that when using the `blocTest` wrapper function, the initial state `0` is not emitted, only the changes that were made during `act`. I am fine with both approaches, so having the initial state always in the state recordings, or leaving it out. Just wanted to mention it. So let's leave the initial state in the recordings, yes?

packages/state_queue_test/lib/src/state_queue_test.dart
wwwdata: Yes, you are always going to get out an array/iterable to assert stuff on anyways. So if you want to test if it is empty you just put it an empty array.
And with only `Iterable` you can then either put in exact states to match or an array of Matchers to further testing on the states. This is very useful if the state is somewhat complex and you don't want to create it entirely for the test. Then you could just use an `isA` matcher and pick some properties to verify

tp: Allowed `Matcher`s inside the `Iterable` in 730e740095822c7c1a489bba736a5fb8925bbd45

packages/state_queue_test/lib/src/state_queue_test.dart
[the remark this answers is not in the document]
tp: > So let's leave the initial state in the recordings, yes?

Agree :)

HenriBeck (approved):

tp:
Published at https://pub.dev/packages/state_queue_test - which still needs some sample in the README

PR #165, opened 2020-11-08
NNBD support

tp:
Flutter version: `1.24.0-10.2.pre`

Based on https://github.com/flutter/flutter/wiki/Experimenting-with-null-safety-in-Flutter


packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -22,10 +22,10 @@ class AuthorizationCredentialAppleID {
     @required this.givenName,
     @required this.familyName,
     @required this.email,
-    @required this.authorizationCode,
+    @required String? authorizationCode,
     @required this.identityToken,
     @required this.state,
-  }) {
+  }) : authorizationCode = authorizationCode ?? '' {
```
tp, line 28: This check we might want to move before the constructor, as it's a little unwieldy here.

packages/sign_in_with_apple/lib/src/exceptions.dart
```diff
@@ -15,39 +15,39 @@ abstract class SignInWithAppleException implements Exception {
     switch (exception.code) {
       case 'not-supported':
         return SignInWithAppleNotSupportedException(
-          message: exception.message,
+          message: exception.message ?? 'no message provided',
```
tp, line 18: Or we also make the message optional all the way through.

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -183,15 +179,15 @@ class SignInWithApple {
...
+  static Future<bool> isAvailable() async {
+    return (await channel.invokeMethod<bool>('isAvailable')) ?? false;
   }
 
   static Future<AuthorizationCredentialAppleID> _signInWithAppleAndroid({
-    @required List<AppleIDAuthorizationScopes> scopes,
-    @required WebAuthenticationOptions webAuthenticationOptions,
-    @required String nonce,
-    @required String state,
+    @required required List<AppleIDAuthorizationScopes> scopes,
```
tp, line 187: I guess these can all be reduced to the single, new `required`, though I don't recall docs on it.

packages/sign_in_with_apple/lib/src/authorization_credential.dart
HenriBeck: An authorization code is also always expected, right?

HenriBeck: Wait, expanding the code below answers the question 😁 

I guess this should move to the caller then as I'm also not sure to which `authorizationCode` the constructor actually refers to here (deducting from NNDB that it is the value from the constructor and not the class)

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -122,33 +109,33 @@ class AuthorizationCredentialPassword {
...
 
 AuthorizationCredentialPassword parseAuthorizationCredentialPassword(
-  Map<dynamic, dynamic> response,
+  Map<dynamic, dynamic>? response,
 ) {
-  if (response['type'] == 'password') {
+  if (response?['type'] == 'password') {
     return AuthorizationCredentialPassword(
-      username: response['username'] as String,
+      username: response!['username'] as String,
```
HenriBeck, line 134: Why do we only need an `!` here and not in the line below?

packages/sign_in_with_apple/lib/src/authorization_request.dart
```diff
@@ -44,9 +44,9 @@ enum AppleIDAuthorizationScopes {
 class AppleIDAuthorizationRequest implements AuthorizationRequest {
   const AppleIDAuthorizationRequest({
     this.scopes = const [],
-    this.nonce,
-    this.state,
-  }) : assert(scopes != null);
+    required this.nonce,
+    required this.state,
```
HenriBeck, line 48:
```suggestion
    this.state,
```

packages/sign_in_with_apple/lib/src/exceptions.dart
```diff
@@ -78,8 +78,8 @@ class UnknownSignInWithAppleException extends PlatformException
 /// An [SignInWithAppleException] which will be thrown in case Sign in with Apple is not supported.
 class SignInWithAppleNotSupportedException implements SignInWithAppleException {
   const SignInWithAppleNotSupportedException({
-    @required this.message,
-  }) : assert(message != null);
+    required this.message,
```
HenriBeck, line 81: Is `required` now a keyword of Dart and you don't need the annotation anymore?

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -203,18 +199,14 @@ class SignInWithApple {
...
-              }
-              return null;
-            })
-            .where((scope) => scope != null)
-            .join(' '),
+        'scope': scopes.map((scope) {
+          switch (scope) {
+            case AppleIDAuthorizationScopes.email:
+              return 'email';
+            case AppleIDAuthorizationScopes.fullName:
```
HenriBeck, line 206: Oh does this mean switch cases are exhaustive when used in combination with enums? 🤯

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -122,33 +109,33 @@ class AuthorizationCredentialPassword {
 
 // ignore_for_file: avoid_as, public_member_api_docs
 AuthorizationCredentialAppleID parseAuthorizationCredentialAppleID(
-  Map<dynamic, dynamic> response,
+  Map<dynamic, dynamic>? response,
```
eernstg, line 112: It might be helpful to require that `response` is not null: In the code below, `response == null` causes the `throw` expression to be evaluated, so the caller might as well ensure that it isn't null (it may be known at the call site).

The null-check in line 116 could then be eliminated (I also don't understand why it would be needed in that line and not in the following lines, cf. line 134), and the two invocations of `operator []` in line 114 and 125 could then omit the `?`.

A similar change could be useful with similar functions, including line 129

packages/sign_in_with_apple/lib/src/authorization_credential.dart
```diff
@@ -171,15 +158,16 @@ AuthorizationCredentialAppleID parseAuthorizationCredentialAppleIDFromDeeplink(
...
       : null;
-  final name = user != null ? user['name'] as Map<String, dynamic> : null;
+  final name = user != null ? user['name'] as Map<String, dynamic>? : null;
 
   return AuthorizationCredentialAppleID(
     authorizationCode: deeplink.queryParameters['code'],
-    email: user != null ? user['email'] as String : null,
-    givenName: name != null ? name['firstName'] as String : null,
-    familyName: name != null ? name['lastName'] as String : null,
+    email: user != null ? user['email'] as String? : null,
```
eernstg, line 168:
```suggestion
    email: user?['email'] as String?,
```

should suffice. Similarly for line 169 and 170.

packages/sign_in_with_apple/lib/src/exceptions.dart
eernstg: Yes, and `required` is stronger: It is an integrated part of the type system. So this change is strongly recommended.

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -183,15 +179,15 @@ class SignInWithApple {
...
 
   static Future<AuthorizationCredentialAppleID> _signInWithAppleAndroid({
-    @required List<AppleIDAuthorizationScopes> scopes,
-    @required WebAuthenticationOptions webAuthenticationOptions,
-    @required String nonce,
-    @required String state,
+    @required required List<AppleIDAuthorizationScopes> scopes,
+    @required required WebAuthenticationOptions webAuthenticationOptions,
+    @required String? nonce,
+    @required String? state,
```
eernstg, line 190: It is true that there is no need to have `@required required`, but there shouldn't actually be _any_ `@required` in code with null safety, because `required` is checked consistently as part of the type system.

Line 189 and 190 should most likely be changed to use `required`, and it may then be possible to keep using the type `String`. (A nullable type for an optional named parameter is possible, but it is most likely a mistake.)

packages/sign_in_with_apple/lib/src/web_authentication_options.dart
```diff
@@ -5,10 +5,9 @@ import 'package:meta/meta.dart';
 /// As described in https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms
 class WebAuthenticationOptions {
   const WebAuthenticationOptions({
-    @required this.clientId,
-    @required this.redirectUri,
-  })  : assert(clientId != null),
-        assert(redirectUri != null);
+    @required required this.clientId,
+    @required required this.redirectUri,
```
eernstg, line 9: Could delete `@required` again, lines 8, 9.

packages/sign_in_with_apple/lib/src/widgets/apple_logo_painter.dart
```diff
@@ -9,8 +9,8 @@ import 'package:flutter/widgets.dart';
 class AppleLogoPainter extends CustomPainter {
   /// Creates an Apple Logo Painter with the provided [color]
   const AppleLogoPainter({
-    @required this.color,
-  }) : assert(color != null);
+    @required required this.color,
```
eernstg, line 12: Could delete `@required`.

packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_builder.dart
```diff
@@ -8,12 +8,10 @@ Widget _fallbackBuilder(BuildContext context) => SizedBox.shrink();
 /// A widget for conditionally rendering UI based on whether or not Sign in with Apple is available
 class SignInWithAppleBuilder extends StatefulWidget {
   SignInWithAppleBuilder({
-    Key key,
-    @required this.builder,
+    Key? key,
+    @required required this.builder,
```
eernstg, line 12: Could delete `@required`.

packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
```diff
@@ -10,19 +10,14 @@ const _appleIconSizeScale = 28 / 44;
 /// https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/
 class SignInWithAppleButton extends StatelessWidget {
   const SignInWithAppleButton({
-    Key key,
-    @required this.onPressed,
+    Key? key,
+    @required required this.onPressed,
```
eernstg, line 14: Could delete `@required`.

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
tp: Removed in 6221538df18cabf47069a1155fb660c09e60de88

packages/sign_in_with_apple/lib/src/authorization_credential.dart
tp: Because after this line (the first `!`), Dart knows that response must be non-`null` if execution continued :)

packages/sign_in_with_apple/lib/src/sign_in_with_apple.dart
tp: > A nullable type for a required named parameter is possible, but it is often a mistake.

In this case it's explicitly desired, as per our preferred coding style we want the caller to pass all parameters, even if they only have `null`.

packages/sign_in_with_apple/lib/src/authorization_request.dart
tp: Right, here we don't want to force the external callers to pass in `null` if they don't want it.

Changed in 18ea5d4a82a94698e57d6cd0bd2f1f6f76e85c4e

packages/sign_in_with_apple/lib/src/exceptions.dart
tp: Will keep the fallback for now, as the optional message seems to be a specialty of `PlatformException` only.

.github/workflows/dart.yml
```diff
@@ -47,3 +40,40 @@ jobs:
...
+        os: [ubuntu-16.04, macos-10.15]
+        package: [sign_in_with_apple]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-java@v1
+        with:
+          java-version: "12.x"
+      - uses: subosito/flutter-action@26bf50084a0c218b5b715a3641c36eac84fd2136 # v1.4.0
+        with:
+          channel: "master" # For now, until new version is released with font loading fix
```
tp, line 58: TODO: Pick a specific version before merge

packages/sign_in_with_apple/example/lib/main.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:io';
 
 import 'package:flutter/material.dart';
+// ignore: import_of_legacy_library_into_null_safe
```
tp, line 4: TODO: Revisit before merge.

(I assume `http` will be updated by then, and then this becomes obsolete)

packages/sign_in_with_apple/lib/src/authorization_credential.dart
tp: Moved the check to the  caller (before calling the constructor), where we can also better log the source of the missing data.

Changed in d45665d0504467c7c372a01c9a8ae66c475b7ad3

HenriBeck (approved):

PR #168, opened 2020-11-23
Migrate remaining packages to NNBD

tp:
(Created a follow up PR to have some cleaner diff on the few changes needed for the remaining packages.)


packages/with_bloc/lib/src/with_bloc.dart
```diff
@@ -107,7 +102,7 @@ class WithBloc<BlocType extends ValueNotifier<StateType>, StateType>
 class WithBlocState<BlocType extends ValueNotifier<StateType>, StateType>
     extends State<WithBloc<BlocType, StateType>> {
   @visibleForTesting
-  BlocType bloc;
+  late BlocType bloc;
```
HenriBeck, line 105: Did you already talk about the usage of `late` in the App? If I'm not mistaken this is a pretty easy escape hatch for the whole code.

I don't think it needs to be changed here but I was just wondering whether or not you would allow the usage in the daily development of the App itself?

HenriBeck (approved):

packages/with_bloc/lib/src/with_bloc.dart
tp: I think for `State` instance fields I see 2 options:
- `late`, basically as we handle them right now (declared var, but then set in `initState`, sometimes "final", sometimes not)
- Set them via the `State`s constructor through `createState`

I think the latter pattern didn't really take hold, but for all the widget which just take an `initial*` argument (and then modify it themselves), it seems like a valid way to get into null-safe territory.

But for many widget related case this seems like the best option (e.g. something animation related that depends on getting `vsync: this`).

PR #182, opened 2021-01-26
don't add to the sink if the taskQueue is closed

agrapine:


packages/state_queue/lib/src/state_queue.dart
```diff
@@ -188,8 +188,10 @@ abstract class StateQueue<T> extends ValueNotifier<T>
   /// of the bloc can not be awaited properly.
   @protected
   void run(StateUpdater<T> updater) {
-    assert(!_taskQueue.isClosed);
-
+    if(_taskQueue.isClosed) {
+      return;
```
HenriBeck, line 192: I would still keep the assert just so this could be caught in tests/development as this would also indicate an issue with the application where something is either disposed to early or not properly cleaned up.

HenriBeck (approved):

packages/state_queue/lib/src/state_queue.dart
agrapine: the assert would be nice if there would be something to do about it. usually the sink gets closed as the bloc gets disposed but it hasn't finished the stack yet.

i would suggest simple log.warn() so you get to see the how it behaves in release mode.

tp: FYI @HenriBeck the problem this is trying to address is this: https://sentry.io/organizations/aym/issues/1868390800/?environment=Production&project=1507839&query=is%3Aunresolved+Cannot+add+event+after+closing&statsPeriod=14d

2 `StateQueue`s are connected via a callback, the whole page (and both blocs) are disposed, but one is still calling the other after it has loaded its data.

This change would silence the error created on the `callee`.

Personally I would rather make the `caller` `StateQueue` aware that is has been disposed (e.g. a private tracking flag), so it can avoid calling its callbacks if that is the case.

If think that would be very much in case with the callbacks we invoke from `State` objects, where we also always check `mounted` first (though that is mostly for our `Provider` usage 😉).

agrapine: i've added `isDisposed` as suggested. this can now be used to stop further runs to be scheduled on the disposed object.

tp: Great, so with `isDiposed` the calling bloc can now avoid that.

I would still prefer to err in case a disposed bloc is called / is calling `run` itself. Just silently doing nothing seems so bad.

Just think about the the `Completer` pattern which would leave one stuck with a `Future` that never resolves. (Sure, in general every caller should handle non-resolving `Future`s somehow (e.g. `timeout`), but in the app code that is never done for internal calls.)

tp: (I just don't think we should add a fix for the original issue at both ends. Because while we can fix the currently known issue properly now still (by adding the `isDisposed` check), the other change would lead to us never again getting an error for these "broken connections")

agrapine: You still get the error in debug console. this way you don't change the behaviour when in release mode.

agrapine: I've removed the early return. the `isDisposed` should suffice.

packages/state_queue/lib/src/state_queue.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:developer' as developer;
```
tp, line 2:
```suggestion
```

packages/state_queue/lib/src/state_queue.dart
```diff
@@ -188,8 +189,8 @@ abstract class StateQueue<T> extends ValueNotifier<T>
   /// of the bloc can not be awaited properly.
   @protected
   void run(StateUpdater<T> updater) {
-    assert(!_taskQueue.isClosed);
-
+    assert(!isDisposed);
+
```
tp, line 193:
```suggestion

```

tp (approved):

tp:
Thank you @agrapine!

Issue #180, opened 2021-01-21
[sign_in_with_apple] Listen for credentialRevokedNotification event


Dave0921:
Add a stream to listen for `credentialRevokedNotification` notification like in this plugin [here](https://github.com/tomgilder/flutter_apple_sign_in/blob/2beebcb0a9e3c62c5f1b27fe2d15b6a9531b55e4/lib/apple_sign_in.dart#L23-L31).

This is to handle the case where during app use, if the user goes into the Settings app and revokes Apple sign-in access to the app, and then returns to the app.

PR #193, opened 2021-03-10
[sign_in_with_apple] Listen for credentialRevokedNotification event

Dave0921:
Add a stream to listen for `ASAuthorizationAppleIDProvider.credentialRevokedNotification` notification. This is to handle the case where the user revokes Apple sign-in access to the app during app use. Related to https://github.com/aboutyou/dart_packages/issues/180.

### DIFF
--- a/packages/sign_in_with_apple/ios/Classes/SignInWithAppleAuthorizationHandler.swift
+++ b/packages/sign_in_with_apple/ios/Classes/SignInWithAppleAuthorizationHandler.swift
@@ -1,0 +1,36 @@
+import Foundation
+import Flutter
+import AuthenticationServices
+
+/*
+ A class that handles the case where during app use, if the user goes into the Settings app and
+ revokes Apple sign-in access.
+ */
+class SignInWithAppleAuthorizationHandler: NSObject, FlutterStreamHandler {
+
+    private var events: FlutterEventSink?
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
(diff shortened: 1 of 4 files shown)


justforworkandstuff:
@tp thinking this would be a great feature to have with the package, is there any concern to not including this feature for the package?

justforworkandstuff:
hey @tp thanks for the prompt reply and appreciate the explanation from your pov. 

I was having scarce knowledge recording sign in with apple and didn't know that apple may communicate directly to the server url once any revoke/delete account action is performed on the apple dashboard itself. 

Originally, my use case was having the client side to listen for the event changes, and then update the server accordingly once any revoke/delete account actions happens. Well, this could work to take off some complexity from the server, but I highly doubt that this approach is a recommended way.

Cheers though.

Artein:
I have iOS app (frontend) and .Net Core Web API (backend). On my backend I implemented server-to-server notifications and successfully delete the user. Similarly on frontend iOS app I want to listen to `ASAuthorizationAppleIDProviderCredentialRevoked` event in case user unlinks AppleId through Settings WHILE my app is running (yes, the app is paused, but can be opened immediately after). 

There is even worse scenario when user logged in on several devices (say, iPhone + iPad) and performs unlink on one of them, while the other is active (it is actually valid case for me).

So, @tp, it is naturally to expect listening of `ASAuthorizationAppleIDProviderCredentialRevoked` in this package to handle such scenarios

PR #194, opened 2021-03-11
Updated to reflect service id is for web

yosus:
Updated to reflect service id is for web

### DIFF
--- a/packages/sign_in_with_apple/README.md
+++ b/packages/sign_in_with_apple/README.md
@@ -72,7 +72,7 @@ If you have change your app's capabilities, you need to fetch the updated provis
 
 #### Create a Service ID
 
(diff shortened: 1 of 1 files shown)


packages/sign_in_with_apple/README.md
```diff
@@ -72,7 +72,7 @@ If you have change your app's capabilities, you need to fetch the updated provis
 
 #### Create a Service ID
 
-Next go to https://developer.apple.com/account/resources/identifiers/list/serviceId and follow these steps:
+Used for website login only. Go to your apple developer page then ["Identifiers"](https://developer.apple.com/account/resources/identifiers/list) and follow these steps:
```
HenriBeck, line 75: I would mention that this is used for Web & **Android** (which ofc works over the web but mentioning this explicitly makes sense here)

PR #209, opened 2021-05-24
Wrap text in fitted box

SaltyPandaStudios:
Wrap the text in a fitted box to ensure localised versions of the string "Sign in with Apple" follow the design guidelines and don't newline when the text is too long


tp:
@SaltyPandaStudios Is there any documentation that scaling is desired? I would hope that all localized texts are picked short enough that they would fit normally.

As I recall there definitely something in the docs about the text and logo height being aligned, I wonder if the better approach wouldn't be to scale the whole button/content?

HenriBeck:
Closing due to inactivity

PR #220, opened 2021-08-09
bug: fixed old Podfile format

### LINKED ISSUE #219: [sign_in_with_apple] Invalid `Podfile` file: no implicit conversion of nil into String on pod install example
When running pod install from the ios folder of the example, the following error is thrown:
`Invalid `Podfile` file: no implicit conversion of nil into String.` 

This is caused by an old podfile generated before flutter 2.0 and can be fixed by regenerating the podfile.

juliansteenbakker:
Fixes #219


HenriBeck (approved):

HenriBeck:
Thanks for the fix

PR #226, opened 2021-08-24
NNBD

agrapine:


HenriBeck:
@agrapine I think Timm already migrated all packages to NNBD

tp:
> @agrapine I think Timm already migrated all packages to NNBD

Indeed, I think we just need to upgrade to the latest published version.

PR #237, opened 2021-09-22
chore(ci): Add building of macos application

HenriBeck:
With 3.1.0, we added support for XCode 13, but this broke the `macOS` build for the `sign_in_with_apple` package.

This package does the following two things:
- Add CI step for building the example `macOS` application
- Remove the handling of the new `ASAuthenticationError.notInteractive` case, which isn't available on macOS


tp:
Is the macOS build failure maybe caused by this line (which assume that the Xcode 13 also uses the macOS 12 SDK to get the new enum case)?

https://github.com/aboutyou/dart_packages/blob/cadf76b57bc0560349291c1d11130682f6956ab1/packages/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift#L110

I didn't find the correct syntax yesterday to check for both the iOS and macOS version, but I suppose that could be improved.

HenriBeck:
> Is the macOS build failure maybe caused by this line (which assume that the Xcode 13 also uses the macOS 12 SDK to get the new enum case)?
> 
> https://github.com/aboutyou/dart_packages/blob/cadf76b57bc0560349291c1d11130682f6956ab1/packages/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift#L110
> 
> I didn't find the correct syntax yesterday to check for both the iOS and macOS version, but I suppose that could be improved.

Yes it is. Afaik the proper syntax would be `#if #available(macOS 10.15, *)` though I'm not 100% sure about the version here

HenriBeck:
@tp any idea how to fix this?

Even an `#available` doesn't work even though the handler is very clearly marked as only being available since iOS 15:
Screenshot 2021-09-22 at 21 23 47

```swift
 if #available(iOS 15.0, *) {
  print(ASAuthorizationError.notInteractive == code)
} else {
  // Fallback on earlier versions
}
```

HenriBeck:
@tp, so removing the case allows all platforms to build successfully; I assume it will just warn everyone using the package during the build process.

packages/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
```diff
@@ -107,11 +107,6 @@ public enum SignInWithAppleError {
                 errorCode = "authorization-error/notHandled"
             case .failed:
                 errorCode = "authorization-error/failed"
-#if swift(>=5.5)
```
HenriBeck, line 110: @tp maybe we also need to do `swift(>=5.5) && os(iOS)`, but this seems in some form wrong tbh 😅

tp: Might be the best approach for now to keep both building.

But I would expect this to fail in a future Xcode version then.

tp:
> @tp any idea how to fix this?
> 
> Even an `#available` doesn't work even though the handler is very clearly marked as only being available since iOS 15…

I think the problem here is that this is used in a "normal `if`", and thus while the condition won't run in practice, the line referencing the unknown enum case is still there and wouldn't get removed / ignored by the compiler.

tp:
Also didn't find a good way to restrict it to the SDK version being used, seems like this is an overall missing feature requested by others as well: https://forums.swift.org/t/do-we-need-something-like-if-available/40349/16

packages/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
tp: Seems best for now, would proceed with this then.

tp (approved):

PR #241, opened 2021-09-23
chore: Add building of the Android App to the CI

HenriBeck:


.github/workflows/dart.yml
```diff
@@ -1,6 +1,10 @@
 name: Dart CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
```
tp, line 6: 👍

tp (approved):

PR #258, opened 2021-10-29
[sign_in_with_apple] Allow optional disabling of SignInWithAppleButton

msisinni:
Making a button's onPressed VoidCallback null is the standard way buttons are disabled in Flutter. Users may have a valid reason for disabling this button.

### DIFF
--- a/packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
+++ b/packages/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart
@@ -11,7 +11,7 @@ const _appleIconSizeScale = 28 / 44;
 class SignInWithAppleButton extends StatelessWidget {
   const SignInWithAppleButton({
     Key? key,
-    required this.onPressed,
+    this.onPressed,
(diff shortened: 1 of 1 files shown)


tp:
> Users may have a valid reason for disabling this button.

Do you have any use-case for this?

---

I can get behind the "`null` handler disables the button" approach, but I would wish that if we make this part of the core package we should provide styling that clearly suggests that the button is not enabled.

But first let's clarify what the intended use-case is.

msisinni:
> > Users may have a valid reason for disabling this button.
> 
> Do you have any use-case for this?
> 
> I can get behind the "`null` handler disables the button" approach, but I would wish that if we make this part of the core package we should provide styling that clearly suggests that the button is not enabled.
> 
> But first let's clarify what the intended use-case is.

I have other social sign in buttons besides this one for Apple. After pressing one of those buttons, I need to disable it and all other buttons while my app communicates with my server. If other social sign in buttons aren't disabled, someone can sign in with one service and then start signing in with another while one is loading.

This button already has different styling for when it is disabled. I wouldn't object to it being looked at more though.

PR #262, opened 2021-11-06
[sign_in_with_apple] Web Support #2: Create platform interface

tp:
…and migrate the existing plugin to use it


packages/sign_in_with_apple/sign_in_with_apple/pubspec.yaml
```diff
@@ -14,6 +14,9 @@ dependencies:
 
   meta: ^1.3.0
 
+  sign_in_with_apple_platform_interface:
+    path: ../sign_in_with_apple_platform_interface # TODO(tp): This must be changed to the published package
```
tp, line 18: This we need to fix before merge by publishing the package and using the published one here

tp:
Published at https://pub.dev/packages/sign_in_with_apple_platform_interface and import changed to point to pub.dev.

packages/sign_in_with_apple/sign_in_with_apple/pubspec.yaml
tp: Done.

packages/sign_in_with_apple/sign_in_with_apple_platform_interface/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
+name: sign_in_with_apple_platform_interface
+description: A common platform interface for the sign_in_with_apple plugin.
+repository: https://github.com/aboutyou/dart_packages/tree/master/packages/sign_in_with_apple/sign_in_with_apple_platform_interface
+version: 1.0.0
```
HenriBeck, line 4: Is there any guideline regarding versioning these platform interface/implementation packages?

E.g., keep all packages on the same version (maybe also just the major version would be enough)?

packages/sign_in_with_apple/sign_in_with_apple_platform_interface/lib/sign_in_with_apple_platform_interface.dart
```diff
@@ -0,0 +1,131 @@
...
+export 'package:sign_in_with_apple_platform_interface/credential_state.dart';
+export 'package:sign_in_with_apple_platform_interface/exceptions.dart';
+export 'package:sign_in_with_apple_platform_interface/nonce.dart';
+export 'package:sign_in_with_apple_platform_interface/web_authentication_options.dart';
+
+import 'method_channel_sign_in_with_apple.dart';
+
+/// The interface that implementations of `sign_in_with_apple` must implement.
+///
+/// Platform implementations should extend this class rather than implement it as `url_launcher`
```
HenriBeck, line 20:
```suggestion
/// Platform implementations should extend this class rather than implement it as `sign_in_with_apple`
```

packages/sign_in_with_apple/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -1,149 +1,35 @@
...
-
-      if (response == null) {
-        throw Exception('getKeychainCredential: Received `null` response');
-      }
-
-      return parseAuthorizationCredentialPassword(response);
-    } on PlatformException catch (exception) {
-      throw SignInWithAppleException.fromPlatformException(exception);
-    }
+  static Future<bool> isAvailable() async {
```
HenriBeck, line 17: Right now we would lose the documentation on these methods (in the Editor and on pub.dev). Should we maybe make the methods non-static and then use `implements SignInWithAppleInterface`?

packages/sign_in_with_apple/sign_in_with_apple_platform_interface/pubspec.yaml
tp: I think the overall guideline is just to be 100% compliant with semver and not have any breaking changes without a major version upgrade.

Because in practice it's expected that the interface is implementation by third party packages as well, and no all (e.g. platform, native mobile, web) would be updated at the same time. So it's important that plugins from different sources are all compatible with the same major version of the interface.

(That's also the reason why there is the `extends` pattern, in which case platform plugins automatically "get the new methods", though in practice they would just throw until it's implemented. But that's by design / the Flutter guidelines.)

packages/sign_in_with_apple/sign_in_with_apple/lib/src/sign_in_with_apple.dart
tp: I shuffled all the docs around in #263 so we get them inline in the editor again.

HenriBeck (approved):

PR #263, opened 2021-11-06
[sign_in_with_apple] Web Support #3: Implement web support

tp:
I opted for the pop-up based flow, so one doesn't loose their "position" in the Flutter app.

The setup works, but is not ideal in my opinion:
- We need to include the JS on the page (always, it only works when initially present on page load)
- For local development you need a HTTPS server (as only those can be added as "redirect URIs" in the service configuration)

I would think we can avoid both drawbacks by changing the flow so that we open our "own" pop-up (just like on Android), and then upon completion we would use a page as `redirectURI` which uses `postMessage` (or similar) so then the credentials back to the opening Flutter website.

While this makes the inclusion in the page a bit lighter (no Apple JS), it will somewhat burden every consumers' backend, as they now need to provide this redirect page that returns the credentials from the pop up.

I would think this is a good first version, and if there is reasonable demand to avoid the SiwA JS we can explore an avenue like the above.

https://user-images.githubusercontent.com/52437/140618966-d57a0815-f69a-4a7a-8172-364bdaaed2a2.mov

TODO:
- [x] Extend the documentation to show what's needed for local plugin development (HTTPS, Apple service config, etc.)


packages/sign_in_with_apple/sign_in_with_apple_web/lib/sign_in_with_apple_web.dart
```diff
@@ -0,0 +1,133 @@
...
+/// Spec: https://developer.apple.com/documentation/sign_in_with_apple/signinresponsei
+@JS()
+@anonymous
+class SignInResponseI {
+  external AuthorizationI get authorization;
+  external UserI? get user;
+}
+
+@JS()
+@anonymous
```
HenriBeck, line 113: @tp the version where we don't depend on the SiwA JS script didn't work?

tp: > @tp the version where we don't depend on the SiwA JS script didn't work?

See the opening comment for details, but in the end not using the JS would impose more implementation effort on each consuming app. Hence I choose this as the default, but the other approach would be a valid addition/refinement.

If we don't use Apple JS, it'd mean we'd have to either redirect ourselves in the current frame (which is a bit bad as we loose the place in the app / have to restart it), or have to open "our own" pop-up, but then the final redirect has to be to a new page which is able communicate with the host app (this is then something the consuming app would have to implement).

I would prefer the "own pop-up" (or `iframe`, if Apple allows it), but didn't want to make it too complicated for consumers to implement.

.github/workflows/dart.yml
```diff
@@ -133,3 +133,22 @@ jobs:
...
+        package:
+          [sign_in_with_apple/sign_in_with_apple]
+        flutter:
+          ["2.5.3"]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: subosito/flutter-action@4389e6cbc6cb8a4b18c628ff96ff90be0e926aa8 # v1.5.3
+        with:
+          flutter-version: ${{ matrix.flutter }}
+      - name: Build Android
```
HenriBeck, line 152:
```suggestion
      - name: Build Web
```

packages/sign_in_with_apple/sign_in_with_apple/example/lib/main.dart
```diff
@@ -63,7 +73,9 @@ class _MyAppState extends State<MyApp> {
                     if (credential.familyName != null)
                       'lastName': credential.familyName!,
                     'useBundleId':
-                        Platform.isIOS || Platform.isMacOS ? 'true' : 'false',
+                        !kIsWeb && (Platform.isIOS || Platform.isMacOS)
```
HenriBeck, line 76: The `Platform` enum is also not available on web?

packages/sign_in_with_apple/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -14,14 +14,61 @@ class SignInWithApple {
...
+  /// A specialized [SignInWithAppleAuthorizationException] is thrown in case of authorization errors, which contains
+  /// further information about the failure.
+  ///
+  /// Throws an [SignInWithAppleNotSupportedException] in case Sign in with Apple is not available (e.g. iOS < 13, macOS < 10.15)
   static Future<AuthorizationCredentialAppleID> getAppleIDCredential({
     required List<AppleIDAuthorizationScopes> scopes,
+
+    /// Optional parameters for web-based authentication flows on non-Apple platforms
+    ///
+    /// This parameter is required on Android.
```
HenriBeck, line 59:
```suggestion
    /// This parameter is required on Android and on the Web.
```

packages/sign_in_with_apple/sign_in_with_apple/lib/src/sign_in_with_apple.dart
```diff
@@ -14,14 +14,61 @@ class SignInWithApple {
...
+  /// - Web
+  ///
+  /// In case Sign in with Apple is not available, the returned Future completes with `false`.
   static Future<bool> isAvailable() async {
     return SignInWithApplePlatform.instance.isAvailable();
   }
 
+  /// Requests an Apple ID credential.
+  ///
+  /// Shows the native UI on Apple's platform and a Chrome Custom Tab on Android.
```
HenriBeck, line 34:
```suggestion
  /// Shows the native UI on Apple's platform, a Chrome Custom Tab on Android, and a popup on Websites.
```

HenriBeck (approved):

packages/sign_in_with_apple/sign_in_with_apple/example/lib/main.dart
tp: No, accession this crashes. There is a similar thing in `dart:html`, but that can only tell you if you're on the web.

Found using the compiler constant as a workaround to only access this when possible. (I guess another option would be to write 2 files with the same interface, for native and web, and then use another conditional import.)

Issue #271, opened 2021-12-15
SignInWithApplePlugin.kt: (159, 29): Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin


nerder:
Since updating flutter to the latest available version in the channel beta we are facing an issue while trying to compile.

```
[✓] Flutter (Channel beta, 2.9.0-0.1.pre, on macOS 11.4 20F71 darwin-x64, locale en-ES)
    • Flutter version 2.9.0-0.1.pre at /Users/nerder/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 8f1f9c10f0 (18 hours ago), 2021-12-14 13:41:48 -0800
    • Engine revision 234aca678a
    • Dart version 2.16.0 (build 2.16.0-80.1.beta)
    • DevTools version 2.9.1
```

```
Running "flutter pub get" in checkin...
Launching lib/main_dev.dart on SM G973F in debug mode...
Running Gradle task 'assembleDevDebug'...
Note: /home/deathcode/.pub-cache/hosted/pub.dartlang.org/share_plus-2.1.4/android/src/main/java/dev/fluttercommunity/plus/share/SharePlusPlugin.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
e: Incompatible classes were found in dependencies. Remove them from the classpath or use '-Xskip-metadata-version-check' to suppress errors
e: /home/deathcode/.gradle/caches/transforms-2/files-2.1/121c56a4eb3b9888b3a925c5a1b06c22/jetified-kotlinx-coroutines-android-1.5.2.jar!/META-INF/kotlinx-coroutines-android.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.5.1, expected version is 1.1.15.
e: /home/deathcode/.gradle/caches/transforms-2/files-2.1/246d7067bcc442dabda444cac586ac0f/jetified-kotlin-stdlib-1.5.31.jar!/META-INF/kotlin-stdlib.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.5.1, expected version is 1.1.15.
[...]

PR #272, opened 2021-12-15
Update kotlin to 1.6.10

deathcoder:
closes https://github.com/aboutyou/dart_packages/issues/271

### DIFF
--- a/packages/sign_in_with_apple/sign_in_with_apple/android/build.gradle
+++ b/packages/sign_in_with_apple/sign_in_with_apple/android/build.gradle
@@ -2,7 +2,7 @@ group 'com.aboutyou.dart_packages.sign_in_with_apple'
 version '1.0-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.3.50'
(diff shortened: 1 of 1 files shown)


tp:
@deathcoder Should we treat this as a breaking change? Meaning if one has ones local project setup to Kotlin `1.3.*` (the default with Flutter 2.5?), will this cause problems when one would update to the latest package with this?

(Currently OOO, but can try this out locally later today.)

tp:
Okay, so the failing CI build seems to answer my previous question.

Then I suppose we'll have to bump the CI build to the point to the upcoming Flutter version, release the plugin with a breaking change and bump the Flutter minimum version in the `pubspec.yaml`.

HenriBeck:
@tp I would suggest only bumping the version officially once the Flutter version is also released to the stable channel.

For now, we could pull in this change into one of our branches, and then for anyone to use it, they would have to point to an exact commit

HenriBeck:
@tp @deathcoder I pulled in the change in our repository (#275) and let's continue the change once Flutter 2.9 is released.

PR #278, opened 2022-01-13
Prefer mavenCentral for Android builds.

asmengistu:
https://blog.gradle.org/jcenter-shutdown

jcenter is shutting down and this can cause Android build failures in some regions immediately, and globally soon.


tp (approved):
LGTM, thanks!

PR #280, opened 2022-01-19
Enable capability for developer accounts

MCYBA:
Add `com.apple.developer.applesignin` (or with create `Runner.entitlements` file) to enabling capability for developer accounts.


HenriBeck:
@MCYBA this is already explained in line https://github.com/aboutyou/dart_packages/pull/280/files#diff-628bde00ccef01ac9c827093c28ae972e80f29636e48141ec98d8179ff549ff9R219

In which case would be an explanation good without using XCode? Right now, I can't think of any use-case 🤔

MCYBA:
> @MCYBA this is already explained in line https://github.com/aboutyou/dart_packages/pull/280/files#diff-628bde00ccef01ac9c827093c28ae972e80f29636e48141ec98d8179ff549ff9R219
> 
> In which case would be an explanation good without using XCode? Right now, I can't think of any use-case thinking

It would be good if the developer doesn't have any Mac and it uses some third-party CI/CD application such as codemagic.

PR #282, opened 2022-01-22
Add tests of success state return in implementation_test

Abhishek01039:


tp:
@Abhishek01039 I think checking that the test helper correctly works in "success" cases is already covered enough in https://github.com/aboutyou/dart_packages/blob/4b296534e055651d76b99101b67f8551e8e8431f/packages/state_queue_test/test/state_queue_test/consumer_test.dart

I suppose the naming could be clearer between the 2 files, but other than that we seem to have everything covered, right?

Abhishek01039:
Hello @tp 
You are right. We have tests both the cases (success and failure)

PR #287, opened 2022-02-19
Bugfix in Flutter v2.11.0 0.0.pre.632 (on master channel)

noordawod:
With most recent version of Flutter, the `onActivityResult()` signature has changed -- the final argument `data` is now non-null. This pr fixes the problem.


HenriBeck:
@noordawod since this breaking change is only on the master channel, we will only merge it once the change comes to the stable channel

noordawod:
> @noordawod since this breaking change is only on the master channel, we will only merge it once the change comes to the stable channel

This is backward-compatible and will run on the stable channel. The change isn't big, and I've tested it on stable.

PR #275, opened 2021-12-25
feat(sign in with apple): Upgrade to kotlin 1.6.10

HenriBeck:

PR #288, opened 2022-02-20
Kotlin Version Error Resolution in Flutter v2.10.x Version - sign_in_with_apple

doch2:
In the current flutter v2.10.x version, if the cotlin version is not more than a certain version, an error occurs and the application does not run normally.
To solve this problem, we updated the Kotiln version of the plug-in and example app and proceeded with the commit.


nerder:
This is a duplicate of https://github.com/aboutyou/dart_packages/pull/275

PR #310, opened 2022-05-18
chore: Add Xcode 13.3.1

### LINKED ISSUE #305: [sign_in_with_apple] `sign_in_with_apple-3.3.0/macos/Classes/SignInWithAppleError.swift:99:13: note: do you want to add missing cases?` when building for macOS
## Description
When building for macOS, you will get the following warning:
```
/Users/nils/.pub-cache/hosted/pub.dartlang.org/sign_in_with_apple-3.3.0/macos/Classes/SignInWithAppleError.swift:99:13: warning: switch must be exhaustive
            switch code {
            ^
/Users/nils/.pub-cache/hosted/pub.dartlang.org/sign_in_with_apple-3.3.0/macos/Classes/SignInWithAppleError.swift:99:13: note: do you want to add missing cases?
            switch code {
            ^
```

## Steps to reproduce
```sh
flutter create test
cd test
flutter pub add sign_in_with_apple
flutter run -d macos
```

### Flutter
Flutter: `2.10.4`
sign_in_with_apple: `3.3.0`
XCode: `13.3.1`

HenriBeck:
Closes #305

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -6,9 +6,6 @@ on:
       - master
   pull_request:
 
-env:
-  DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
-
 jobs:
   test:
     name: Test ${{ matrix.package }} on ${{ matrix.os }}
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -82,6 +82,8 @@ jobs:
       - uses: subosito/flutter-action@4389e6cbc6cb8a4b18c628ff96ff90be0e926aa8 # v1.5.3
         with:
           flutter-version: ${{ matrix.flutter }}
+      - name: Flutter doctor (version check)
+        run: flutter doctor
       - name: Build iOS
         run: flutter build ios --no-codesign
         working-directory: packages/${{ matrix.package }}/example
@@ -142,6 +144,8 @@ jobs:
           flutter-version: ${{ matrix.flutter }}
       - name: Enable macOS Desktop Integration
         run: flutter config --enable-macos-desktop
+      - name: Flutter doctor (version check)
+        run: flutter doctor
       - name: Build macOS
         run: flutter build macos
         working-directory: packages/${{ matrix.package }}/example
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -109,13 +109,20 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-12]
+        os: [macos-11, macos-12]
         package:
           [sign_in_with_apple/sign_in_with_apple]
         xcode:
           ["12.5.1", "13.0", "13.3.1"]
         flutter:
           ["2.8.1", "2.10.2"]
+        exclude:
+        - os: macos-11
+          xcode: 13.3.1
+        - os: macos-12
+          xcode: 12.5.1
+        - os: macos-12
+          xcode: 13.0
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-java@v1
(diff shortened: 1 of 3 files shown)


tp (approved):
Wondering whether we still need the `DEVELOPER_DIR` at the start of the file 🤔
Don't recall what we added that for specifically, but maybe it's part of the bootstrapping (and not the actual build, which should of course use the directory of the desired Xcode)

.github/workflows/dart.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         package:
           [sign_in_with_apple/sign_in_with_apple]
         xcode:
-          ["12.5.1", "13.0"]
+          ["12.5.1", "13.0", "13.3.1"]
```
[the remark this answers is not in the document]
tp: But that has only the most recent versions then (https://github.com/actions/virtual-environments/blob/main/images/macos/macos-12-Readme.md#xcode). 

Wondering whether there is still some need to test with an other Xcode toolchain…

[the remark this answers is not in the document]
HenriBeck: @tp I think it makes sense to test different versions especially because of issues like #305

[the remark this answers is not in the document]
tp: Totally. Cool fix with the exclude, didn't know that yet :)

.github/workflows/dart.yml
```diff
@@ -58,13 +58,20 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-11]
+        os: [macos-12]
```
tp, line 61:
```suggestion
        os: [macos-11, macos-12]
```

Right? As the excludes are already in place, but the OS is not selected.

tp (approved):
LGTM, with the macOS 11 reintroduction for the iOS build still pending.

tp:
> Wondering whether we still need the `DEVELOPER_DIR` at the start of the file 🤔


I get the feeling that this still has an impact, maybe we never switched versions correctly?

Otherwise I can't currently explain how #305 has that build error in Xcode and we don't see that in CI.


Will push a test commit to output the selected version to find that out.

tp:
> Will push a test commit to output the selected version to find that out.

Ah, turns out those errors also exist in CI, but don't fail the build as they're "just" warnings.

CleanShot 2022-05-19 at 09 51 45@2x

.github/workflows/dart.yml
tp: Added.

tp (approved):
Okay, finally happy with this :)

.github/workflows/dart.yml
```diff
@@ -6,8 +6,9 @@ on:
       - master
   pull_request:
 
-env:
-  DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
+concurrency:
```
HenriBeck, line 9: 👍

packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
```diff
@@ -107,8 +107,9 @@ public enum SignInWithAppleError {
                 errorCode = "authorization-error/notHandled"
             case .failed:
                 errorCode = "authorization-error/failed"
-#if os(iOS) && swift(>=5.5)
-            // new case since Xcode 13, but only on iOS SDK at the moment
+#if os(iOS) && swift(>=5.5) || os(macOS) && swift(>=5.5.1)
```
HenriBeck, line 110:
```suggestion
#if (os(iOS) && swift(>=5.5)) || (os(macOS) && swift(>=5.5.1))
```

Just to make it clear in regards to priority?

tp: Sure, why not :)

Issue #162, opened 2020-10-29
Sign In cancelled when app go to background


Danaru87:
**Flutter** : 1.22.0
**Android**: 10
**sign_in_with_apple**: ^2.5.4

We create an application using the Apple Sign In plugin.

When the web view asks for an Apple ID and password, two-factor authentication sends a code via SMS.

But, going to the SMS application to read the code, going back to the Flutter application causes an exception of type: SignInWithAppleAuthorizationError (AuthorizationErrorCode.canceled, The user closed the Custom Tab).

tp:
@Danaru87 Can you provide a screen recording of how the switch towards the SMS app happens?

I would only expect this on resource constrained devices when the browser activity gets closed.

Danaru87:
Flutter: 1.22.3
sign_in_with_apple: ^2.5.4
Device: Motorola One Vision
Android: 10

video:


![bug apple sign in android](https://user-images.githubusercontent.com/5873203/98663420-604a5100-2349-11eb-9d13-b4e27572e316.gif)


error message in console:
```
E/flutter (22882): [ERROR:flutter/lib/ui/ui_dart_state.cc(177)] Unhandled Exception: SignInWithAppleAuthorizationError(AuthorizationErrorCode.canceled, The user closed the Custom Tab)
E/flutter (22882): #0      SignInWithApple._signInWithAppleAndroid (package:sign_in_with_apple/src/sign_in_with_apple.dart:239:7)
E/flutter (22882): <asynchronous suspension>
E/flutter (22882): #1      SignInWithApple.getAppleIDCredential (package:sign_in_with_apple/src/sign_in_with_apple.dart:107:14)
E/flutter (22882): #2      _ProviderButtonsWidget.build.<anonymous closure> (package:spheery/provider_choice/provider_choice_screen.dart:95:58)
E/flutter (22882): #3      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:993:19)
E/flutter (22882): #4      _InkResponseState.build.<anonymous closure> (package:flutter/src/material/ink_well.dart:1111:38)
E/flutter (22882): #5      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:183:24)
E/flutter (22882): #6      TapGestureRecognizer.handleTapUp (package:flutter/src/gestures/tap.dart:598:11)
[...]

PR #345, opened 2022-12-11
#162: Ensure Chrome Custom Tab does not disappear…

tp:
…when switching away and back from the app during Sign in with Apple

### DIFF
--- a/packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
+++ b/packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.3.0
+
+- Android: Fixes an issue where the Chrome Custom Tab would disappear when the user "left" the app (e.g. using the app switcher) ([#162](https://github.com/aboutyou/dart_packages/issues/162))
+  - Beware that your app's `launchMode` affects the specific behavior of the Chrome Custom Tab. For more information see the README.
+
 ## 4.2.0
 
 - Remove `jcenter` repository for the Android package
--- a/packages/sign_in_with_apple/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
+++ b/packages/sign_in_with_apple/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
@@ -91,7 +91,6 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
 
         val builder = CustomTabsIntent.Builder();
         val customTabsIntent = builder.build();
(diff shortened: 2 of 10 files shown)


packages/sign_in_with_apple/sign_in_with_apple/android/src/main/kotlin/com/aboutyou/dart_packages/sign_in_with_apple/SignInWithApplePlugin.kt
```diff
@@ -91,7 +91,6 @@ public class SignInWithApplePlugin: FlutterPlugin, MethodCallHandler, ActivityAw
 
         val builder = CustomTabsIntent.Builder();
         val customTabsIntent = builder.build();
-        customTabsIntent.intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
```
HenriBeck, line 94: So the fix is to keep the activity on the history?

I can't exactly recall why we added this in the initial version, do you?

tp: Tested both deep links (SiwA and normal app (main activity) ones), and they would with both `launchMode`s close the CCT. For the in-app deep link the deep link handler would be called inside the app. 

So this seems fine as far as I can tell.

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+### 5.0.0
```
tp, line 1: After the extensive testing, I wouldn't deem this a breaking change, as in both cases (`launchMode`s) it just works better.

So if testing from the reports in #162 is also fine, I would consider releasing this as just `4.3.0`.

What do you think @HenriBeck ?

packages/sign_in_with_apple/sign_in_with_apple/example/README.md
```diff
@@ -13,3 +13,25 @@ flutter config --enable-macos-desktop
 # cd macos && bundle exec pod install && cd -
 flutter run -d macOS
 ```
+
+## Test Android deep links
```
tp, line 17: Extended the sample app a bit so we can easily test the deep links with it.

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
HenriBeck: Should we add the docs also the the README?

HenriBeck (approved):

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
tp: Good point. I'll adapt the version number and add the `launchMode` info into the README for Android.

PR #375, opened 2023-05-09
[sign_in_with_apple] Add support for Gradle 8+

### LINKED ISSUE #374: Android native pluging doesn't support AGP 8.0
The native Android plugin of this package doesn't support Android gradle Plugin 8.0, in 8.0 the package name of a library or application should be defined in build.gradle in the app module in the android block as a namespace instead of specifying it in the package attribute in AndroidManifest.xml, however, that might break compatibility with gradle versions older than 7.0, but gradle 8.0 will be required sooner or later, I think all the packages that have native plugin should update android gradle plugin to 8.0 so people start to migrate their build system to 8.0 to significantly improve the build system of android and how the app is packaged, All applications and libraries that use plugin have gradle version less than 8.0 can't migrate unless they update it manually but that is not a good solution since after any upgrade or clean the changes will reverted back

[Fore more information](https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace)

davidmartos96:
## Description
Fixes #374

Hello, currently the Apple Sign In plugin doesn't build when Gradle 8+ is used. That is because the property `namespace` is mandatory.

* I updated the Android example project.
* I included the namespace property with backwards compatibility


tp:
Thanks @davidmartos96!

I can give this spin tonight to test on device.

packages/sign_in_with_apple/sign_in_with_apple/pubspec.lock
```diff
@@ -144,58 +153,58 @@ packages:
...
       name: vector_math
-      url: "https://pub.dartlang.org"
+      sha256: "80b3257d1492ce4d091729e3a67a60407d227c27241d6927be0130c98e741803"
+      url: "https://pub.dev"
     source: hosted
-    version: "2.1.1"
+    version: "2.1.4"
 sdks:
-  dart: ">=2.14.0 <3.0.0"
+  dart: ">=2.18.0 <3.0.0"
```
tp, line 209: Should be fine as the package's lock files are not used for the importing applications, but I'll see whether this would cause an issue when still developing with the version specified in

https://github.com/aboutyou/dart_packages/blob/master/packages/sign_in_with_apple/sign_in_with_apple/pubspec.yaml#L8

In which case we should probably keep this "downgraded".

davidmartos96: Ok I see. What should we change then? Only the `dart: ">=2.18.0 <3.0.0"` into `>=2.14.0` ?

tp: Locally I had to use at least Flutter 3.7 in order to be able to build with the current Android Studio (tried 2.10 and 3.3 without any luck).

Thus I bumped the example's required to the most recent stable version, as I think people will only upgrade this in conjunction with Flutter (+ Gradle) anyway (as I don't expect any feature updates on the SiwA package soon).

Given that Flutter 3 has been released for 1 year tomorrow, I think we can safely assume that consumers have migrated any potential breaking changes, and thus should be quick to update to 3.7.x soon as well.

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.4.0
+
+- Support Gradle 8 ([@davidmartos96](https://github.com/davidmartos96) in [#375](https://github.com/aboutyou/dart_packages/pull/375))
+  - Requires at least Flutter 3.7.12 which brings full Gradle 8 support
```
tp, line 4: TBD if we should then bump to version 5.0.0. I think it's fine an consumers will just pick the highest package they can support.

Or maybe we find a way to keep this fully compatible with Flutter 2/3.x, but I was unable to get that working locally.

davidmartos96: Are you trying in the example app? Or a new project?
Maybe this is relevant? https://docs.flutter.dev/release/breaking-changes/android-java-gradle-migration-guide

davidmartos96: Which particular 2.x and 3.x versions are not working? I assume 2.8.0 since it's the minimum stated in the pubspec

cpboyd:
I had this error:
```
Execution failed for task ':sign_in_with_apple:compileDebugKotlin'.
> 'compileDebugJavaWithJavac' task (current target is 1.8) and 'compileDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
  Consider using JVM toolchain: https://kotl.in/gradle/jvm/toolchain
```

adding this (which is in the current Flutter template for `build.gradle`) fixed the error:

```groovy
android {
...
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }

    kotlinOptions {
        jvmTarget = '1.8'
    }
...
}
```

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
tp: > Are you trying in the example app? Or a new project?

I was looking at the example app.

I tried 2.8, 3.0, and 3.3 (all in their latest patch releases).

But I suppose it might be related to my locally upgraded Android Studio + Java version.

Maybe there is a way to keep better backwards compatibility, I'll check this against the migrations we did in the production app itself.

davidmartos96: I'll try 2.8 and 3.3 with previous Android Studio (Eel). That could be the problem indeed.
I'm not sure there is a backwards compatibility fix when using AS Flamingo, given that Flutter has this in their home page (https://docs.flutter.dev/release/breaking-changes/android-java-gradle-migration-guide)

davidmartos96: @tp  I managed to run it correctly with Flutter 2.8 and 3.3.
What I did:

* Install Android Studio Electric Eel (https://developer.android.com/studio/archive)
* Remove the recently added >=3.7 constraint
*  Keep the `androidx.browser:browser` dependency in 1.4.0. Before this PR it was 1.3.0 and latest is 1.5.0 but 1.5.0 requires android SDK 33. Flutter 2.8 and 3.3 by default, when installing, use SDK 31. I'm Flutter < 3.7 supports SDK 33. 
Changelog for the browser library: https://developer.android.com/jetpack/androidx/releases/browser

davidmartos96: I run the example app by opening the `example/android` folder with this Android Studio. I believe a special flag for `flutter run` is needed to provide a different android studio installation, so I found easier to do `flutter pub get` and run from Android Studio directly

packages/sign_in_with_apple/sign_in_with_apple/pubspec.yaml
```diff
@@ -1,12 +1,12 @@
...
-version: 4.3.0
+version: 4.4.0
 homepage: https://github.com/aboutyou/dart_packages/tree/master/packages/sign_in_with_apple
 repository: https://github.com/aboutyou/dart_packages
 
 environment:
-  sdk: ">=2.12.0 <3.0.0"
-  flutter: ">=2.8.0"
+  sdk: ">=2.18.5 <3.0.0"
+  flutter: ">=3.7.6 <4.0.0"
```
HenriBeck, line 9: This would need to be released as v5 if we bump the flutter version to 3.x right?

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
tp: @davidmartos96 Oh nice, good to know that it'd still work with a legacy Android Studio.

Question is to we really want to put in the extra effort to keep this supported, or maybe take this as a good time to require at least e.g. Flutter 3.7.6 which most people should be using by now anyway. Unless someone chimes in here with a specific backwards-compatibility requirement, I think it would be beneficial to ease our support burden here.

packages/sign_in_with_apple/sign_in_with_apple/pubspec.yaml
tp: Yeah, had the same thought. I suppose that would make sense then with a clear breaking change dropping the legacy versions.

Even though it shouldn't cause any trouble (as client could not upgrade to this version), it's probably best to consciously make consumers aware of this upgrade anyway.

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
davidmartos96: SDK 33 will be a requirement in August. I see 2 options.
1. Update the constraint to 3.7.6 now and update the browser dependency to 1.5.0
2. Keep the 2.8.0 constraint with browser 1.4.0 and in a few months update it, since SDK 33 (Flutter 3.7) will be a requirement.

davidmartos96: Source https://support.google.com/googleplay/android-developer/answer/11926878

davidmartos96: Yes, same thoughts.
Should the browser dependency with updated to 1.5.0 on Android? 1.3.0 is from December 2, 2020

packages/sign_in_with_apple/sign_in_with_apple/pubspec.yaml
tp: Let's do v5 then :)

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
tp: It is already, right?

https://github.com/aboutyou/dart_packages/blob/71741b933bc3d0846af1178f7ee41e0f0e959b55/packages/sign_in_with_apple/sign_in_with_apple/android/build.gradle#L58

tp:
@HenriBeck This now looks good from my point of view.

packages/sign_in_with_apple/sign_in_with_apple/CHANGELOG.md
davidmartos96: @tp 😅 Totally right, testing it with the old AS confused me haha

PR changes LGTM!

HenriBeck (approved):

tp:
This is now published as `v5.0.0` https://pub.dev/packages/sign_in_with_apple/versions/5.0.0

Thank you @davidmartos96!

davidmartos96:
@tp Great, thank you!

PR #403, opened 2024-02-23
Migrate to `package:web`

### LINKED ISSUE #401: Migrate to `package:web` to support WASM
The short version:
- Migrate usage of dart:html to package:web
- Migrate usage of dart:js, dart:js_util, and package:js to dart:js_interop

This will allow compiling to WASM engine instead of just JavaScript on web.

Full migration guide: https://dart.dev/interop/js-interop/package-web

tp:
Fixes #401


HenriBeck (approved):

bartekpacia:
Why was min Flutter version set to 3.19.1 instead of 3.19.0?

bartekpacia:
> Is that causing any issue on your end?

A minor one, a [macOS VM image I use](https://github.com/cirruslabs/macos-image-templates/pkgs/container/macos-sonoma-xcode) comes with Flutter 3.19.0. Therefore I'd appreciate if minimum version was downgraded to 3.19.0.

If you agree with this proposal, then I will contribute this tiny change.

> test with that version as well

The only change between 3.19.0 and 3.19.1 is [this](https://github.com/flutter/flutter/wiki/Hotfixes-to-the-Stable-Channel#3191-february-21-2024) and it isn't relevant to this plugin.

PR #448, opened 2024-11-12
Deeplink query paramerters user details validation improvment

kariot:
Added validation to the user object received via the deep link. It already checks for the presence of the user object, but it could still be empty in sign-in scenarios and interrupt the sign-in flow due to JSON decode error.


kariot:
Hi @tp,

I've noticed the issue where I was receiving an empty user object—meaning the user key is present, but it has no associated value. I was able to reproduce this by signing up from an iOS device and then signing in from an Android device, which seems to cause a crash.

Issue #311, opened 2022-05-19
Handle "app transferred" state


tp:
Look's like [SignInWithAppleAvailablePlugin.swift](https://github.com/aboutyou/dart_packages/blob/a38f85e6035bf8029721e27a080ea19195c8fac1/packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleAvailablePlugin.swift#L83) is not handling the [transferred](https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidprovider/credentialstate/transferred) case of [CredentialState](https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidprovider/credentialstate).

_Originally posted by @erf in https://github.com/aboutyou/dart_packages/issues/305#issuecomment-1131229161_

Should be handled here: https://github.com/aboutyou/dart_packages/blob/e7f2e4b820acb5882eb9714a5ad8812734544392/packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleAvailablePlugin.swift#L84

PR #455, opened 2024-12-11
Support `transferred` credential state

tp:
Relates to #311, #454


tp:
Should be fine, but want to test in CI after https://github.com/aboutyou/dart_packages/pull/456 is merged.

@marcotta For the time being you could point your app to this branch. But I think I can do a (preview) release tomorrow as well :)

PR #457, opened 2025-01-17
fix: add missing switch cases and update compiler check

juliansteenbakker:
Fixes warning during build about switch case not being exhaustive

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -130,13 +130,17 @@ jobs:
       matrix:
         os: [macos-13, macos-14]
         package: [sign_in_with_apple/sign_in_with_apple]
-        xcode: ["14.3.1", "15.2"]
+        xcode: ["14.3.1", "15.2", "16.2"]
         flutter: ["3.19.1"]
         exclude:
           - os: macos-13
             xcode: 14.3.1
           - os: macos-14
             xcode: 14.3.1
+          - os: macos-13
+            xcode: 16.2
+          - os: macos-14
+            xcode: 16.2
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-java@v1
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -68,11 +68,15 @@ jobs:
             xcode: 15.2
           - os: macos-13
             xcode: 16.1
+          - os: macos-13
+            xcode: 16.2
           - os: macos-14
             xcode: 14.3.1
           - os: macos-14
             xcode: 16.1
             flutter: 3.19.1
+          - os: macos-14
+            xcode: 16.2
           - os: macos-15
             xcode: 15.2
           - os: macos-15
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -61,7 +61,7 @@ jobs:
       matrix:
         os: [macos-13, macos-14, macos-15]
         package: [sign_in_with_apple/sign_in_with_apple]
-        xcode: ["14.3.1", "15.2", "16.1"]
+        xcode: ["14.3.1", "15.2", "16.1", "16.2"]
         flutter: ["3.19.1", "3.24.3"]
         exclude:
           - os: macos-13
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -128,7 +128,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-13, macos-14]
+        os: [macos-13, macos-14, macos-15]
         package: [sign_in_with_apple/sign_in_with_apple]
         xcode: ["14.3.1", "15.2", "16.2"]
         flutter: ["3.19.1"]
@@ -141,6 +141,10 @@ jobs:
             xcode: 16.2
           - os: macos-14
             xcode: 16.2
+          - os: macos-15
+            xcode: 15.2
+          - os: macos-15
+            xcode: 14.3.1
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-java@v1
--- a/packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
+++ b/packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
@@ -107,15 +107,21 @@ public enum SignInWithAppleError {
                 errorCode = "authorization-error/notHandled"
             case .failed:
                 errorCode = "authorization-error/failed"
-#if (os(iOS) && swift(>=5.5)) || (os(macOS) && swift(>=5.5.1))
+#if compiler(>=5.5)
             // new case since Xcode 13, arrived earlier in iOS
-            // use https://swiftly.dev/swift-versions to match Swift to Xcode versions (as this is in practice driven by the OS SDK, not Swift version)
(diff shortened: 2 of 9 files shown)


packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
```diff
@@ -107,15 +107,21 @@ public enum SignInWithAppleError {
...
             case .notInteractive:
                 errorCode = "authorization-error/notInteractive"
 #endif
 #if compiler(>=6.0)
             case .matchedExcludedCredential:
                 errorCode = "authorization-error/matchedExcludedCredential"
+#endif
+#if compiler(>=6.0.3)
+            case .credentialImport:
+                errorCode = "authorization-error/credentialImport"
```
tp, line 122: We should map these on the Dart side, as in https://github.com/aboutyou/dart_packages/blob/63b49af2e8cef437a2b4e2bff5adbd5dbe95a86a/packages/sign_in_with_apple/sign_in_with_apple_platform_interface/lib/exceptions.dart#L42

Now I also noticed that the already existing case (in Swift) of `matchedExcludedCredential` seems to be missing there.

.github/workflows/dart.yml
```diff
@@ -129,6 +131,8 @@ jobs:
         exclude:
           - os: macos-13
             xcode: 14.3.1
+          - os: macos-14
+            xcode: 14.3.1
```
tp, line 135: Now it's excluded on both versions, so this would not run.

juliansteenbakker: I think something went wrong while rebasing onto master. Let me try that again

.github/workflows/dart.yml
```diff
@@ -61,15 +61,17 @@ jobs:
       matrix:
         os: [macos-13, macos-14, macos-15]
         package: [sign_in_with_apple/sign_in_with_apple]
-        xcode: ["14.3.1", "15.2", "16.0"]
+        xcode: ["14.3.1", "15.2", "16.1"]
```
tp, line 64: Personally I would like to use at least 16.1 and 16.2 (where the new case was added and what you check with `compiler(>=6.0.3)`, right?).

I don't think we can expect users (and their CI) to always use the very latest Xcode version shortly after release, so we have to be sure to support the currently expected ones.

juliansteenbakker: These changes were not part of my original PR, but changes already merged in https://github.com/aboutyou/dart_packages/pull/456.

I have removed these changes from my PR, so now it should be correct. 

As for the check, i think it would be good to also check for 16.2, so if wanted, i can add that to my PR.

packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
juliansteenbakker: I have added the missing matchedExcludedCredential as wel as the other exceptions to the dart class.

.github/workflows/dart.yml
tp: > As for the check, i think it would be good to also check for 16.2, so if wanted, i can add that to my PR.

Yes, that's what I would also suggest, given that the change you added (with that specific compiler version) will only be included in 16.2.

packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
```diff
@@ -107,15 +107,21 @@ public enum SignInWithAppleError {
                 errorCode = "authorization-error/notHandled"
             case .failed:
                 errorCode = "authorization-error/failed"
-#if (os(iOS) && swift(>=5.5)) || (os(macOS) && swift(>=5.5.1))
+#if compiler(>=5.5)
```
tp, line 110: Was this change needed? The old comment refers to this being shipped in different versions for iOS and macOS.

But overall I would not mind to change this, as we only support Xcode 14.3.1 (with Swift 5.8), so this will work on either OS with this check and that macOS version.

Just generally of the "don't fix what's not broken" camp here, as we don't know if there is not some user relying on this behavior because of a certain old Xcode version (though by now there might be 3 other changes that would've long prevented that one from using the latest version of this package, as we can not test every single combination).

.github/workflows/dart.yml
juliansteenbakker: I have added 16.2 for both macos and ios build.

packages/sign_in_with_apple/sign_in_with_apple/darwin/Classes/SignInWithAppleError.swift
juliansteenbakker: I have reverted the change

tp:
> I have applied your improvements @tp. I hope we can get this merged soon 😄

Thanks @juliansteenbakker. I'll take another look :)

.github/workflows/dart.yml
```diff
@@ -126,13 +130,17 @@ jobs:
       matrix:
         os: [macos-13, macos-14]
```
tp, line 131: Did you mean to add `macos-15` here as well? (And exclude the old version combinations, just like iOS)?

Because right now the new Xcode is added, but then excluded on macOS 13 and 14, which are all of the current options.

packages/sign_in_with_apple/sign_in_with_apple_platform_interface/lib/exceptions.dart
```diff
@@ -49,6 +49,21 @@ abstract class SignInWithAppleException implements Exception {
...
+          message: exception.message ?? 'no message provided',
+        );
+      case 'authorization-error/credentialExport':
+        return SignInWithAppleAuthorizationException(
+          code: AuthorizationErrorCode.credentialExport,
+          message: exception.message ?? 'no message provided',
+        );
+      case 'authorization-error/matchedExcludedCredential':
+        return SignInWithAppleAuthorizationException(
+          code: AuthorizationErrorCode.credentialExport,
```
tp, line 64:
```suggestion
          code: AuthorizationErrorCode.matchedExcludedCredential,
```

packages/sign_in_with_apple/sign_in_with_apple_platform_interface/lib/exceptions.dart
```diff
@@ -114,6 +129,15 @@ enum AuthorizationErrorCode {
...
   /// The authorization attempt failed for an unknown reason.
   unknown,
+
+  /// The credential export request failed.
+  credentialExport,
+
+  /// The credential import request failed.
+  credentialImport,
+
+  /// This error should only be returned when specifying @c excludedCredentials on a public key credential registration request.
```
tp, line 139: What does `@c` mean here?

tp (commented):
Sorry for the late reply, just got back from traveling yesterday.

I think some minor issues slipped in here, but when those are fixed it should be good to go.

packages/sign_in_with_apple/sign_in_with_apple_platform_interface/lib/exceptions.dart
tp: Hmm, weird, this is also similar in the official docs, but I wonder why this case is treated separately from the others?

CleanShot 2025-04-05 at 14 47 24@2x

tp (approved):

tp:
This has been released in `v7.0.0`: https://pub.dev/packages/sign_in_with_apple/versions/7.0.0

PR #462, opened 2025-03-18
Update dependencies and add textWidget parameter to SignInWithAppleButton

IndigoSoftwares21:
Description:
- Updated project dependencies to their latest compatible versions.
- Added a new textWidget parameter to the SignInWithAppleButton component, allowing custom text widgets instead of the default label.
- Ensured backward compatibility by maintaining the default behavior when textWidget is not provided.


IndigoSoftwares21:
Hello @tp thank you for your response, I used a text widget, because the current font style and sizing did not match my app's style/theme.

We can leave a comment in the docs to let user's know the recommended font size.

I will exclude my changes to the `pubspec.yaml` and `pubspec.lock` files

tp (approved):

tp:
Thanks. This is fine with me.

I wonder how we don't offer a localized version of the "Sign in with Apple" text yet.

Maybe we did only do it inside our apps, but there is a default translation for "Sign in with Apple" for all languages Apple support, so developers should use that.

Maybe the blocker was that it would require a certain localization library as a dependency, which would make it a no-go, but maybe also Flutter change in the meantime and we could offer this now.

What I would love then would be to have a builder instead, where the developer can pass a `Widget` that gets the "default text".

But this is all beyond the scope of this one, and we can start putting this in first.

Thanks @IndigoSoftwares21.

tp:
Is there any useful test we should duplicate for this functionality? Maybe just a simple rendering one (no need to for a golden test), that passes a widget like `textWidget: Text('adsf')` and then asserts on that being used?

Dzivo:
This will help us come closer to our designed button

PR #463, opened 2025-03-24
feat(ui): make `onPressed` nullable to support disabled state

BriceFab:
This change allows the `onPressed` callback in `SignInWithAppleButton` to be nullable.

Making it nullable enables use cases where the button needs to be disabled — for example:
- While waiting for async operations (loading)
- When Apple Sign-In is not available on the current platform
- To reflect validation state or form readiness

This improves flexibility and aligns with standard Flutter button behavior (e.g. `TextButton`, `ElevatedButton`).


tp:
Offering something like this indeed seems nice. 
(Alternatively in many projects we also use buttons with `FutureOr<void> Function() onTap`, so the button can go in and out of the loading state for its own behavior itself.)

Right now we could use `null` to disable the button, but I think we'll have to take care of styling it accordingly as well, right? Because otherwise one might just use `() {}` instead of `null` (although it'd be nice to stick to "Flutter"'s convention here).

BriceFab:
@tp using null for onPressed is the proper way to disable a button in Flutter, and that’s also how native Flutter buttons like TextButton, ElevatedButton, and CupertinoButton behave.

Providing an empty callback () {} doesn’t trigger the same result:
	•	The button remains enabled (visually and functionally)
	•	You lose any built-in styling or feedback for disabled state
	•	You can’t rely on it for accessibility purposes

tp:
@BriceFab I agree that this would be a nice addition.

Does this already adapt the styling of the box for the button then?

If so, we should just add a quick golden, and then I think we could ship this.

If the styling is not yet ("nicely") adapted we should take care of that in one swoop.

BriceFab:
@tp Actually, since this is based on CupertinoButton, the disabled state is already handled when onPressed == null.

As per Flutter docs:

“If null, the button will be disabled.”
→ https://api.flutter.dev/flutter/cupertino/CupertinoButton/enabled.html and https://api.flutter.dev/flutter/cupertino/CupertinoButton/disabledColor.html

Also confirmed here:
→ [StackOverflow - “How do I disable a button in Flutter?”](https://stackoverflow.com/questions/49351648/how-do-i-disable-a-button-in-flutter)

So we don’t need to add extra styling logic — CupertinoButton already dims by default.

That said, it might be a nice enhancement to allow customizing the disabled appearance (e.g. via a disabledColor - https://api.flutter.dev/flutter/cupertino/CupertinoButton/disabledColor.html) directly in the SignInWithAppleButton widget.
This would give more flexibility without breaking Cupertino behavior but not the most important for this PR.

https://github.com/aboutyou/dart_packages/blob/master/packages/sign_in_with_apple/sign_in_with_apple/lib/src/widgets/sign_in_with_apple_button.dart#L162

tp:
@BriceFab Okay thanks. I worried that our styling was overriding the base behavior.

But then this seems good. I can check it out locally later and then also add the appropriate golden test to check that this works properly in light and dark mode etc.

tp:
Apart from the black button the disabled state of the others does not feel clear to me (in isolation), but the stock Cupertino buttons have similar issues unless one uses an "active" color like blue for the font or background, which then gets clearly "toned down".

tp:
This has been released in `v7.0.1`: https://pub.dev/packages/sign_in_with_apple/versions/7.0.1

PR #466, opened 2025-04-05
Run Apple platform CI tests on Xcode 16.3

tp:


tp:
Still blocked by https://github.com/actions/runner-images/issues/11916
-> https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md

Maybe we can also switch to an explicit matrix like this one:
https://github.com/firebase/firebase-ios-sdk/pull/14651/files/23140b6d1ff3077056409294cc9a12153cde714b

tp:
> Maybe we can also switch to an explicit matrix for macOS/Xcode like this one

I won't be getting to this right now, but definitely something to keep in mind if the matrix configuration grows.

HenriBeck (approved):

PR #471, opened 2025-06-22
[sign_in_with_apple] ci: Add SPM build for iOS

tp:
Follow up from https://github.com/aboutyou/dart_packages/pull/470

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -106,6 +106,57 @@ jobs:
           OTHER_SWIFT_FLAGS: "-warnings-as-errors"
           DEVELOPER_DIR: /Applications/Xcode_${{ matrix.xcode }}.app/Contents/Developer
 
+  build-ios-spm:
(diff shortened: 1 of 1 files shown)


.github/workflows/dart.yml
```diff
@@ -106,6 +106,57 @@ jobs:
...
+        run: flutter doctor
+      - name: Enable SPM support
+        run:  flutter config --enable-swift-package-manager
+        working-directory: packages/${{ matrix.package }}/example
+      - name: List dependency config before (debug)
+        run: |
+          less ios/Runner.xcodeproj/project.pbxproj | grep Pods || echo ""
+          less ios/Runner.xcodeproj/project.pbxproj | grep FlutterGeneratedPluginSwiftPackage || echo ""
+        working-directory: packages/${{ matrix.package }}/example
+      - name: Remove Cocoapods from project
```
HenriBeck, line 143: I guess also other Flutter projects still support cocoapods or how is this looking there?

tp: Yeah, either way I would like to test both for now.

And I think since SPM is not yet the default one, it makes sense to keep the CP project and just migrate it in CI.

But if need be we can of course add a second clean SPM one (hopefully this can be avoided and we just switch to all-on SPM soon enough).

PR #478, opened 2025-12-11
Bump min version to Flutter 3.35

tp:
In order to work with Flutter 3.35 a few update to the CI and Gradle scripts were needed, as well as minor code fixes, and the new `dart format`.


packages/sign_in_with_apple/sign_in_with_apple/example/pubspec.yaml
```diff
@@ -5,7 +5,7 @@ version: 1.0.0+1
 
 environment:
   sdk: ^3.3.0
-  flutter: ">=3.19.1 <4.0.0"
+  flutter: ">=3.35.0 <4.0.0"
```
tp: Probably should bump this across the board now, so we don't need to special case the examples for CI. It's been out for long enough I suppose.

